PR #2, opened 2017-01-30
enable usage of openigtlink install-type files

dkuegler:
This commit modifies the include statements, so it is compatible with the installed version of openigtlink (as compared to only the built version). This is due to the fact, that OpenIGTLink installs header files to different locations in the install "distribution", specifically, the igtlutil subfolder is just installed into the base include folder.

compatibility witth installed version of OpenIGTLink


adamrankin:
Hi David, thanks for the pull requests.

I don't have the time to deal with them immediately, as I am fixing up the infrastructure so that we have a green dashboard.

Once that is finished I will take a look!

dkuegler:
Sorry, I commited this before I saw the SetupForDevelopment script.

PR #4, opened 2017-02-01
re #1154 adding IP output on server startup

adamrankin:


lassoan:
I think this commit comment is invalid "re #1154 adding IP output on server startup" because of the missing colon. This would be the correct version: "re #1154: adding IP output on server startup"
Do you use setupfordevelopment.sh? It should catch this.

adamrankin:
The colon isn't necessary.

See [https://app.assembla.com/spaces/plus/subversion/commits/list](https://app.assembla.com/spaces/plus/subversion/commits/list)

PR #11, opened 2017-02-08
Explicitly adding ITK libraries to VolumeReconstructor

dzenanz:


src/PlusVolumeReconstruction/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ SET (PlusVolumeReconstruction_LIBS
   vtkRenderingFreeType
   vtkRendering${VTK_RENDERING_BACKEND}
   vtkRenderingVolume${VTK_RENDERING_BACKEND}
+  ${ITK_LIBRARIES}
```
adamrankin: Could you specify only the libraries needed, instead of a broad include?

lassoan:
Thank you for the quick fix. I would prefer to restore the build ASAP, because the latest stable Plus doesn't work with SlicerIGT PlusRemote. So, I merge this now and @dzenanz can later touch up the library list.

adamrankin:
Closing this in favour of #12 which links only the necessary dependency.

PR #12, opened 2017-02-09
re #1148 linking to necessary ITK dependency

adamrankin:
Cleaning up code

### DIFF
--- a/src/PlusVolumeReconstruction/CMakeLists.txt
+++ b/src/PlusVolumeReconstruction/CMakeLists.txt
@@ -34,12 +34,13 @@ SET (PlusVolumeReconstruction_LIBS
   vtkRenderingFreeType
   vtkRendering${VTK_RENDERING_BACKEND}
   vtkRenderingVolume${VTK_RENDERING_BACKEND}
+  ITKIOImageBase
   )
 
 GENERATE_EXPORT_DIRECTIVE_FILE(vtkPlusVolumeReconstruction)
 ADD_LIBRARY(vtkPlusVolumeReconstruction ${PlusVolumeReconstruction_SRCS} ${PlusVolumeReconstruction_HDRS})
-target_include_directories(vtkPlusVolumeReconstruction PUBLIC ${PlusVolumeReconstruction_INCLUDE_DIRS} )
-TARGET_LINK_LIBRARIES(  vtkPlusVolumeReconstruction  ${PlusVolumeReconstruction_LIBS} )
+target_include_directories(vtkPlusVolumeReconstruction PUBLIC ${PlusVolumeReconstruction_INCLUDE_DIRS})
+TARGET_LINK_LIBRARIES(vtkPlusVolumeReconstruction ${PlusVolumeReconstruction_LIBS})
 
 # --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


adamrankin:
Is there a list variable for ITK IO targets?

dzenanz:
Unfortunately, no. You have to manually list all the IO formats you want
linked. You can find the complete list here
<https://itk.org/Doxygen/html/classitk_1_1ImageIOBase.html>, by clicking on
the Inheritance Diagram.

Dženan Zukić, PhD, Senior R&D Engineer, Kitware (Carrboro, N.C.)

adamrankin:
Obvious question, does the VTK reader have a similar target construction? If not, does it support enough file formats to be reasonable?

adamrankin:
Ah, it has vtkIOImage at least...

dzenanz:
vtkIOImage probably only supports .vtk. And VTK has a separate reader for each format, e.g. vtkPNGReader, vtkBMPReader etc and a switch statement to select between them.

adamrankin:
Ok, how about a CMake loop that scans all ITK targets for "IO" in them, and links to those targets? We may catch some extra targets, but at least it's not ITK_LIBRARIES...

lassoan:
Earlier we only supported nrrd and vtk and nobody complained. It's fine to just list a few file formats that we think people may want to use and we can add support for more, if requested.

adamrankin:
This is for mask loading though, some people may want BMP or PNG as common paint tools output these formats.

lassoan:
Sure, we can add support for bmp, png, jpg. No need for any fancy auto-explore mechanism. ITK's IO plugin infrastructure is fragile enough already (it's easy to forget about IO factory registration, etc).

lassoan:
@dzenanz Is there a test that uses mask and fails if mask file loading fails?

adamrankin:
So far, I have this:

# Scan ITK_LIBRARIES for targets with "IO" in them
SET(_itk_io_libraries)
FOREACH(_lib ${ITK_LIBRARIES})
  IF(${_lib} MATCHES "IO")
    LIST(APPEND _itk_io_libraries ${_lib})
  ENDIF()
ENDFOREACH()

It produces a reasonable list.

adamrankin:
If we only want those file formats, then I'd vote we use VTK to keep our ITK dependencies to a minimum.

lassoan:
I agree. VTK's IO is much simpler and more robust (no IO factory, plugins, no surprises at runtime), and fulfills our basic needs.

adamrankin:
Does the ITK system scan files for headers, or is it also based on file extension?

adamrankin:
@dzenanz What do you think, are the formats supported by VTK enough?

dzenanz:
Well originally I had a code version which was using only vtkPNGReader, anticipating problems with lots of ITK IO dependencies. I was surprised when I didn't need to modify CMakeLists.txt for ITK to work! So I went ahead and submitted the version with ITK only.

dzenanz:
#13 has only vtkPNGReader. To support merging of 3D scans, a reader supporting a 3D image format should be added. I don't have any such data to test with.

PR #32, opened 2017-02-20
re #1170 fix the proper deduction of the git directory

dkuegler:
now using git rev-parse --git-dir


adamrankin:
We implemented the same fix! Closing.

PR #34, opened 2017-03-13
Added a freeze/unfreeze call to the Capistrano video device after an 'unknow…

jperdomo23:
…n USB error' or a 'Lost probe synchronization' error is caught. This fixes a bug where the video stream does not unfreeze after one of these errors is caught.

### DIFF
--- a/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
+++ b/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
@@ -1263,12 +1263,16 @@ PlusStatus vtkPlusCapistranoVideoSource::WaitForFrame()
   break;
   case USB_NOTSEQ:
     LOG_ERROR("Lost Probe Synchronization. Please check probe cables and restart.");
+	FreezeDevice(true);
+	FreezeDevice(false);
     break;
   case USB_STOPPED:
     LOG_ERROR("USB: Stopped. Check probe and restart.");
     break;
   default:
     LOG_ERROR("USB: Unknown USB error: "<<usbErrorCode);
+	FreezeDevice(true);
+	FreezeDevice(false);
     break;
   }
 


dzenanz:
@lassoan @adamrankin Queens' server requires password to see build log, and Robarts' is timing out.

adamrankin:
Yes, apologies for the pull request builds. They are not an accurate indication.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -1263,12 +1263,16 @@ PlusStatus vtkPlusCapistranoVideoSource::WaitForFrame()
   break;
   case USB_NOTSEQ:
     LOG_ERROR("Lost Probe Synchronization. Please check probe cables and restart.");
+	FreezeDevice(true);
```
lassoan: I would recommend to add a comment here: why this is needed, what does it do/what happens if this freeze/unfreeze is not done.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -1263,12 +1263,16 @@ PlusStatus vtkPlusCapistranoVideoSource::WaitForFrame()
...
     LOG_ERROR("Lost Probe Synchronization. Please check probe cables and restart.");
+	FreezeDevice(true);
+	FreezeDevice(false);
     break;
   case USB_STOPPED:
     LOG_ERROR("USB: Stopped. Check probe and restart.");
     break;
   default:
     LOG_ERROR("USB: Unknown USB error: "<<usbErrorCode);
+	FreezeDevice(true);
```
lassoan: I would recommend to add a comment here: why this is needed, what does it do/what happens if this freeze/unfreeze is not done.

PR #36, opened 2017-03-30
re #1063 Publicly exposing freeze setter and status

dzenanz:


lassoan:
You may consider using standard naming conventions. For example GetFreeze(), SetFreeze(bool), FreezeOn(), FreezeOff().

PR #40, opened 2017-04-26
fixes #1181 Support building PlusLib as part of Slicer's superbuild

dzenanz:
ExternalProject_Add's CMAKE_CACHE_ARGS should have these:
  -DPLUS_USE_SYSTEM_ZLIB:BOOL=ON
  -DZLIB_INCLUDE_DIR:PATH=${ZLIB_INCLUDE_DIR}
  -DZLIB_LIBRARY:FILEPATH=${ZLIB_LIBRARY}
  -DITK_NO_IO_FACTORY_REGISTER_MANAGER:BOOL=ON

PLUS_USE_SYSTEM_ZLIB should be used in cases when ITK's and VTK's internal zlib build has been disabled.

ITK_NO_IO_FACTORY_REGISTER_MANAGER should be used in all the programs and libraries using itk::ImageFileReader and itk::ImageFileWriter classes. This makes sure there is only a single ITK I/O factory manager.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,6 +12,9 @@ ENDIF()
 # Try to find ITK and include its settings (otherwise complain)
 FIND_PACKAGE (ITK REQUIRED PATHS ${ITK_DIR} NO_DEFAULT_PATH)
 IF(ITK_FOUND)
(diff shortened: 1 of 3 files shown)


lassoan:
Thank you!

PR #41, opened 2017-04-27
re #1182 replacing direct string compare with case insensitive compare

adamrankin:
Automated tests pass, but could someone take a look?

### DIFF
--- a/src/PlusCommon/PlusCommon.cxx
+++ b/src/PlusCommon/PlusCommon.cxx
@@ -15,6 +15,10 @@ See License.txt for details.
 #include <vtkXMLDataElement.h>
 #include <vtksys/SystemTools.hxx>
 
+// STL includes
+#include <algorithm>
+#include <string>
+
 //-------------------------------------------------------
 PlusTransformName::PlusTransformName()
 {
@@ -596,6 +600,12 @@ namespace
   {
     return ::tolower(a) == ::tolower(b);
   }
+
+  //----------------------------------------------------------------------------
+  bool icompare_pred_w(wchar_t a, wchar_t b)
+  {
+    return ::towlower(a) == ::towlower(b);
+  }
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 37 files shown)


src/PlusUsSimulator/PlusSpatialModel.cxx
```diff
@@ -518,14 +521,13 @@ PlusStatus PlusSpatialModel::UpdateModelFile()
 }
 
 //-----------------------------------------------------------------------------
-void PlusSpatialModel::SetModelFile(const char* modelFile)
+void PlusSpatialModel::SetModelFile(const std::string& modelFile)
 {
-  std::string oldModelFile = this->ModelFile;
-  this->ModelFile = modelFile ? modelFile : "";
-  if (this->ModelFile.compare(oldModelFile) != 0)
+  if (!PlusCommon::IsEqualInsensitive(modelFile, this->ModelFile))
```
lassoan: file names are case-sensitive on non-Windows systems, I would keep case-sensitive compare

src/PlusCommon/PlusCommon.h
```diff
@@ -47,13 +47,13 @@ enum PlusImagingMode
 
 /* Define case insensitive string compare for Windows. */
 #if defined( _WIN32 ) && !defined(__CYGWIN__)
-#  if defined(__BORLANDC__)
-#    define STRCASECMP stricmp
-#  else
-#    define STRCASECMP _stricmp
-#  endif
+  #if defined(__BORLANDC__)
+    #define STRCASECMP stricmp
```
lassoan: Keep only one way of case-insensitive string compare. I'm OK with the new method but then remove the STRCASECMP macro.

src/PlusCommon/PlusCommon.cxx
```diff
@@ -612,6 +622,41 @@ bool vtkPlusCommonExport PlusCommon::IsEqualInsensitive(std::string const& a, st
 }
 
 //----------------------------------------------------------------------------
+bool vtkPlusCommonExport PlusCommon::IsEqualInsensitive(std::wstring const& a, std::wstring const& b)
```
lassoan: Have you tried if you see a difference in loading time with the original version and after this pull request?
Try to load a header with at least a few ten thousand transforms.

src/PlusCommon/PlusCommon.h
adamrankin: There was no actual change here, just my auto-formatting changing the indent.

lassoan: I know. The issue is that there are multiple case-insensitive string comparison implementation in PlusCommon for us to use, maintain, document. We should keep only one.

adamrankin: Right, gotcha. I view the STRCASECMP for usage with C style strings and new methods for C++ strings. I suppose compilers will automatically construct C++ from C. Still replace?

lassoan: As we move towards std::string, we should probably retire STRCASECMP.
It is mostly used for XML attribute value checks, such as 

`if (toolDataElement->GetAttribute("Type") != NULL && STRCASECMP(toolDataElement->GetAttribute("Type"), "Tool") != 0)`

If we add a helper method for that then the code will be more compact and we can control at one place how this comparison is implemented.

For the rest of the STRCASECMP we need to check, but I guess we should be able to replace them with the new method.

src/PlusCommon/PlusCommon.cxx
adamrankin: 0.296s vs. 0.261s for 23399 frames.

src/PlusCommon/PlusCommon.h
adamrankin: Implementing this cleanup, do not accept PR until next changes.

PR #43, opened 2017-05-03
addresses #1185 use write_basic_package_version_file instead of custo…

dkuegler:
…m Version file


dkuegler:
cf. Pull request 42
since this pull request compiles fine for me, I would like to get Access to the jenkins Logs, how do I do that? I do not have a user name...

lassoan:
Thanks a lot, your changes look good and will merge soon.
@adamrankin Do you think you can fix the automatic builds or we should disable them for now?

adamrankin:
I would disable them. The problems arise from the volume reconstruction tests having different results on different platforms. I am unfamiliar with that code.

lassoan:
If it's only due to the VR tests (that are kind of expected to have minor differences on different computers due to floating-point computations) then we could disable them for these automatic tests.

lassoan:
👍 Thank you for your contribution

PR #44, opened 2017-05-05
re #1187 support for VS2017 and Qt5.7

markasselin:
Fix include directory for generated Qt headers under VS2017 Community
and Qt5.7.


lassoan:
👍 Thank you!

PR #46, opened 2017-05-09
re #1190 fix trying to install PDB files in a static build

dkuegler:
so for my build it was really helpful to have the pdbs in the install directory and thusly i just copied them. It seems I just keep running into situations, where they do not exist (Linux, static build). Sorry for the trouble I have caused.

I have checked Release and RelWithDebInfo Configurations, but I do not currently have a Debug VTK and ITK build. I will fix that shortly, but for now a quick fix of the current issue #1190.


dzenanz:
With this patch, the solution gets generated without issues. My debug build is under way.

dzenanz:
The build completed successfully. Thanks David!

dzenanz:
@lassoan @adamrankin Can someone merge?

lassoan:
The same pdb adding script is repeated several times and it is not that short and simple anymore.
@dkuegler if you have time, it would be nice if you could define a CMake macro or function for this and use that single line everywhere.

dkuegler:
Good idea. Will do.

Vom Smartphone gesendet

David Kügler
Fachgebiet Graphisch-Interaktive Systeme
Technische Universität Darmstadt
Tel: +49 6151 155-672
david.kuegler@gris.tu-darmstadt.de

Am 10.05.2017 12:46 vorm. schrieb Andras Lasso <notifications@github.com>:

The same pdb adding script is repeated several times and it is not that short and simple anymore.
@dkuegler<https://github.com/dkuegler> if you have time, it would be nice if you could define a CMake macro or function for this and use that single line everywhere.

PR #49, opened 2017-05-11
re #1187: Fixed CMake configuration error with CMake 3.6

BenChurch:


src/PlusWidgets/CMakeLists.txt
```diff
@@ -40,7 +40,9 @@ SET(${PROJECT_NAME}_INCLUDE_DIRS
   ${CMAKE_CURRENT_BINARY_DIR}
   ${CMAKE_CURRENT_SOURCE_DIR}
   )
-IF(CMAKE_VERSION VERSION_GREATER_EQUAL "3.8")
+# If CMake 3.7 is the minimum version then the next line can be
```
adamrankin: Why does the previous version fail? 3.6 should not be >= than 3.8?

lassoan: VERSION_GREATER_EQUAL is introduced in CMake 3.7.
In the build instructions, CMake 3.7.2 version is described. I'm wondering if we should make it the minimum required CMake version in the top-level CMakeLists.txt.

adamrankin: Yes, I would like that. 3.7.2 seems to be very stable and all the features we need.

lassoan: Is there anything that is not in 3.6? Just asking because Linux distros etc. are typically lagging behind. If we require 3.7.2 then users must download&install CMake from the website instead of just use what's installed already.

adamrankin: Ah this makes sense... Ok, let's maintain support for 3.6

PR #50, opened 2017-05-16
re #1198: Added command line tool which takes an mha sequence file as…

BenChurch:
… input and the image processing pipeline of vtkPlusTransverseProcessEnhancer on it, saving intermediate and output tracked image sequences to mha files.


adamrankin:
Please add a doxygen page describing how to use the enhancer application.

BenChurch:
Ok, I think I've done it. Is that what you are looking for @adamrankin ?

adamrankin:
Great, talk to Andras to find a good place to link to it from the main document tree!

PR #51, opened 2017-05-17
re #1194 Adding initial WinProbe support

dzenanz:
This PR only adds a basic level of support for WinProbe ultrasound system. In current version of the WinProbeSDK, some of the functions are not properly exported in x64 version, so a few function calls are commented out (Voltage, ADCfrequency).

Currently the test fails, probably because it cannot find the .net DLLs. I tried adding WORKING_DIRECTORY argument but it didn't help. Running the test from VisualStudio works, as long as I set working directory to _PlusBuild-bin\bin\Debug_.

Feedback is welcome!


dzenanz:
RelWithDebInfo version is not working either, a feedback regarding that would be welcome @dkuegler

dkuegler:
Works fine for me. Can you Plesse post the errors?

Vom Smartphone gesendet

David Kügler
Fachgebiet Graphisch-Interaktive Systeme
Technische Universität Darmstadt
Tel: +49 6151 155-672
david.kuegler@gris.tu-darmstadt.de

Am 17.05.2017 7:43 nachm. schrieb Dženan Zukić <notifications@github.com>:

RelWithDebInfo version is not working either, a feedback regarding that would be welcome @dkuegler<https://github.com/dkuegler>

dzenanz:
---------------------------
vtkWinProbeVideoSourceTest.exe - System Error
---------------------------
The program can't start because UltraVisionManagedDll.dll is missing from your computer. Try reinstalling the program to fix this problem. 
---------------------------
OK   
---------------------------

My CMake code copies UltraVisionManagedDll.dll and other DLLs from WinProbeSDK to bin folder, but only for Debug and Release variants. How to accomplish the same with other configurations?

adamrankin:
Perhaps a series of macros would be in order to standardise plus data
collection cmake.

dzenanz:
I copied Capistrano wrapper implementation and mostly followed it, occasionally taking inspiration from Interson. I suspect that Capistrano and Interson would not compile/run properly in RelWithDebInfo and MinSizeRel configurations.

dkuegler:
I currently have no access to my build, so I can have a look at this tomorrow.

Since up to this point no use of cmake macros or functions were in use, I thought this was some kind of policy, but I do think they would increase the redability of the code.

Unfortunately, I only use the Ascension Tracker and the OpenIGTLink interfaces, so I have no experience with other interfaces. I am willing to help implement some cmake modules though. But they should be well defined. I would also advocate to properly define what is included in the different Build Types.

I think RelWithDebInfo should always include .PDB files on Windows platforms (which I implemented in #1184). Before that I contributed cmake code, that created proper install folders and "distributions" that can be shared by people using the same development environment.

dzenanz:
This is the output of configure step with latest commit:
```
1>  Right? way: C:/Dev/Plus64d/bin/$(Configuration)
1>  Wrong way: C:/Dev/Plus64d/bin/Debug
```
Is the rest of Plus build system modifying those paths? How to prevent it? @adamrankin @lassoan

lassoan:
There is no need to mess with the working directory to find DLLs. Just make sure the DLL is copied to binary directory for each configuration (Debug, Release, MinSizeRel, RelWithDebInfo). This consumes some extra disk space, but simpler, cleaner, and more robust than finding DLLs by manipulating PATH environment variable or changing work directories.

dkuegler:
That is definitely not based on anything I have touched.

dkuegler:
... But I do have an idea on how to fix it....

dkuegler:
As commented before, this is independent of any change I contributed and should be solved by copying the relevant DLLs into the respective dirs. Using Working Directories has many pitfalls including that suddenly other DLLs for PlusLib are used, than you suppose...

dzenanz:
What is the default working directory for tests? Some WinProbeSDK XML configuration files need to be in the working directory, so I should copy them there too.

lassoan:
Configuration file search location is stored in PlusConfig.xml. Location of additional config files should be relative to the main config file location. See for example how NDI .rom files are found.

dzenanz:
WinProbeSDK XML files are not Plus-format configuration files, but rather some WinProbe-specific configuration, e.g. description of different type of probes. I think that current directory is searched for them.

lassoan:
This is very sommon - see MicronTracker.ini files, NDI Polaris .rom files, etc. Use the same mechanism as those devices.

dzenanz:
Hey @adamrankin @lassoan, I think this PR is ready for merge.

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1194,6 +1195,57 @@ IF(PLUS_USE_CAPISTRANO_VIDEO)
...
+  SET(PlusDataCollection_SRCS ${PlusDataCollection_SRCS}
+    ${WinProbe_SRCS}
+    )
+
+  SET(PlusDataCollection_INCLUDE_DIRS ${PlusDataCollection_INCLUDE_DIRS}     
+    ${WINPROBESDK_DIR}
+    ${CMAKE_CURRENT_SOURCE_DIR}/WinProbe
+    )
+
+  FILE(GLOB WinProbeXMLFiles "${WINPROBESDK_DIR}/xml/*.xml")
```
adamrankin: For some reason I have this feeling that globbing files is not best practice. Would it be better to enumerate the XML files involved?

If I'm crazy, then disregard this comment.

dzenanz: There is half a dozen XML files, and some 20-ish DLLs. Enumerating them by hand would not only be cumbersome, but also prone to change with each new version of the SDK.

adamrankin: Fair enough.

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1194,6 +1195,57 @@ IF(PLUS_USE_CAPISTRANO_VIDEO)
...
+  FILE(COPY ${WinProbeXMLFiles} DESTINATION ${CMAKE_BINARY_DIR})
```
adamrankin: Highly recommend using PLUS_EXECUTABLE_OUTPUT_DIR (I think, double check the exact variable name)

adamrankin:
Could you run the PlusLib CTest suite?

BuildAndTest.bat/.sh in PlusLib-bin

dzenanz:
I am running it.

src/PlusDataCollection/CMakeLists.txt
lassoan: There are many drawbacks of globbing (look at CMake documentation for details), so although it's slightly inconvenient, it's usually still better to enumerate all files by hand.
Globbing files from a third-party directory is somewhat less of an issue, so if you really want to do it, then I think it's acceptable (we went back and forth on this for other SDK files and ended up globbing for some of them).

dzenanz:
Once ready, merge together with the [accompanying PR](https://github.com/PlusToolkit/PlusBuild/pull/2) for PlusBuild.

src/PlusDataCollection/CMakeLists.txt
dzenanz: This SDK is also currently under development, so changes are likely as we experiment with it. I would much rather keep the globbing approach, at least for now.

src/PlusDataCollection/CMakeLists.txt
dzenanz: If I copy the XML files there, running CTest from command line fails. And working directory path gets modified somehow ([commit](https://github.com/PlusToolkit/PlusLib/pull/51/commits/c363a2a1e798120596a95a13998e368e799b79cd)), so I couldn't just rely on WORKING_DIRECTORY argument for the test.

lassoan: Please, do not use working directory or have any assumptions about the working directory. All configuration files must be found relative to the config file location (see how it is done for NDI Polaris, MicronTracker, etc. devices).

dzenanz:
Thanks Adam!

src/PlusDataCollection/CMakeLists.txt
dzenanz: I don't have control over where from these .xml files are read. I will talk about this with our contact people at WinProbe, and see whether this is configurable.

lassoan: Thanks for the clarification. Yes, they should allow you to specify xml file location.

PR #55, opened 2017-05-24
re #1196: Initial commit to create OpticalMarkerTracking device…

markasselin:
re #1196: Initial commit to create OpticalMarkerTracking device using aruco.  Source files include initial changes based off of vtkPlusIntelRealSenseTracker.

This pull request is in tandem with a pull request for PlusBuild that enables building of aruco - which OpticalMarkerTracking depends on.


src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1703,6 +1705,37 @@ ENDIF()
...
+  SET(PlusDataCollection_HDRS ${PlusDataCollection_HDRS}
+    ${OpticalMarkerTracking_HDRS}
+    )
+
+  SET(PlusDataCollection_SRCS ${PlusDataCollection_SRCS}
+    ${OpticalMarkerTracking_SRCS}
+    )
+
+  SET(PlusDataCollection_LIBS ${PlusDataCollection_LIBS}
+    OpticalMarkerTracking
```
adamrankin: This doesn't seem right. Is there an external library for OpticalMarkerTracking? This seems like it should be aruco and OpenCV

PR #62, opened 2017-06-30
re #1061 USDigital encoders don't require serial numbers

dzenanz:


src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.h
```diff
@@ -14,155 +14,159 @@ See License.txt for details.
...
-  PlusStatus InternalStopRecording();
 
- 
 private:  // Variables.
-  bool                                            coreXY;
-  long                                            NumberOfEncoders; // Number of connected A2 Encoders
-  long                                            COMPort;
-  long                                            BaudRate;
+    bool coreXY;
+    bool configurationViaSerialNumbers = false;
```
lassoan: Either initialize all variables here or none of them.

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.h
```diff
@@ -146,11 +146,11 @@ class vtkPlusDataCollectionExport vtkPlusUSDigitalEncodersTracker : public vtkPl
 
 
 private:  // Variables.
-    bool coreXY;
+    bool coreXY = false;
```
adamrankin: If you're making minor fixes to the class, would you consider moving all variables initialized in the header to the constructor's constructor list?

No biggie if you choose not to, but it would be appreciated.

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.h
```diff
@@ -146,11 +146,11 @@ class vtkPlusDataCollectionExport vtkPlusUSDigitalEncodersTracker : public vtkPl
...
 private:  // Variables.
-    bool coreXY;
+    bool coreXY = false;
     bool configurationViaSerialNumbers = false;
-    long NumberOfEncoders; // Number of connected A2 Encoders
-    long COMPort;
-    long BaudRate;
+    long NumberOfEncoders = 0; // Number of connected A2 Encoders
+    long COMPort = 0;
+    long BaudRate = 9600;
```
adamrankin: Also, I know we can recite these in our sleep, but would you consider seeing if there's an enum to re-use (can move to PlusCommon if there's another device/class using the same values)?

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.h
dzenanz: The simple variables are initialized at place of declaration, the ones requiring function calls in the constructor. Initializing simple variables in the constructor requires more typing, makes maintenance slightly harder, and requires more effort to find out its default value (as it is separated from the declaration).

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.h
dzenanz: The most similar other device is PlusCivcoBrachyStepper, as they both use USDigital's DEIDrv32. But merging common parts into a common ancestor class would be some work, with a question: would that make sense?

adamrankin: Not the whole class, just the baud rate enum.

dzenanz: As I removed it in a later commit, this should be good to go?

adamrankin: Yup!

dzenanz:
@adamrankin @lassoan merge?

dzenanz:
Thanks!

PR #63, opened 2017-07-12
re #1208:  Updated OptiTrack support to use NatNet SDK instead of deprecated TrackingTools.

markasselin:


markasselin:
Thanks for reminding me.  I've sent the pull request for PlusBuild.

lassoan:
Thanks! Do you have an update for PlusApp as well (to update packaging)?

PR #64, opened 2017-07-12
re #1194 Fixing unfreeze method

dzenanz:


dzenanz:
Thanks for a quick merge @adamrankin !

PR #65, opened 2017-07-18
Improvements to Serial device and US Digital encoders

dzenanz:


src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -170,6 +184,70 @@ PlusStatus vtkPlusGenericSerialDevice::InternalUpdate()
   return PLUS_SUCCESS;
 }
 
+PlusStatus vtkPlusGenericSerialDevice::SetDTR(bool OnOff)
```
lassoan: don't use capital letters for local variables (OnOff->onOff)

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -170,6 +184,70 @@ PlusStatus vtkPlusGenericSerialDevice::InternalUpdate()
...
 }
 
+PlusStatus vtkPlusGenericSerialDevice::SetDTR(bool OnOff)
+{
+  // Either update or send commands - but not simultaneously
+  PlusLockGuard<vtkPlusRecursiveCriticalSection> updateMutexGuardedLock(this->Mutex);
+
+  PlusStatus retval;
+
+  if (OnOff == DTR)
```
lassoan: replace DTR by this->DTR to make it clear that it is a member var

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -170,6 +184,70 @@ PlusStatus vtkPlusGenericSerialDevice::InternalUpdate()
...
+  {
+    return PLUS_SUCCESS; //already the desired value
+  }
+  else if (OnOff == true)
+  {
+    retval = this->Serial->SetDTR();
```
lassoan: Instead of SetDTR() and ClearDTR() have a single method SetDTR(bool). It would make the code much shorter and simpler.

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -246,8 +324,16 @@ PlusStatus vtkPlusGenericSerialDevice::ReceiveResponse(std::string& textReceived
       if (vtkPlusAccurateTimer::GetSystemTime() - startTime > this->MaximumReplyDurationSec)
       {
         // waiting time expired
-        LOG_ERROR("Failed to read complete line from serial device. Received: " << textReceived);
-        return PLUS_FAIL;
+        if (textReceived.empty())
```
lassoan: Make the behavior configurable. Add an option "AcceptReply":
* REQUIRE_LINE_ENDING (default): current behavior
* REQUIRE_NOT_EMPTY: the behavior that you implemented in this pull request
* ANY: response either empty or not empty is returns with success

dzenanz:
@lassoan @adamrankin Can you re-check the additional commits?

src/Documentation/UserManual/DeviceGenericSerial.dox
```diff
@@ -6,7 +6,8 @@ It does not collect tracking or image data, only allows sending and receiving da
 using OpenIGTLink commands (see \ref PlusServerCommands).
 
 A (customizable) newline string is appended after each sent command and the device listens
-for reponse that is terminated by the same newline string.
+for reponse that is terminated by the same newline string by default.
+The default can be modified by specifying "acceptReply" parameter of ReplyTermination type.
```
adamrankin: Upper camel case for exposed parameters.

src/PlusDataCollection/PlusSerialLine.cxx
```diff
@@ -235,3 +235,47 @@ unsigned int SerialLine::GetNumberOfBytesAvailableForReading() const
...
+  if (EscapeCommFunction(CommHandle, dwFunc))
+  {
+    return PLUS_SUCCESS;
+  }
+  else
+  {
+    return PLUS_FAIL;
+  }
+#else
+  LOG_ERROR("SerialLine::SetDTR() is only implemented on Windows");
```
adamrankin: Linux implementation does not seem complicated. Could you add it?

See http://xanthium.in/Controlling-RTS-and-DTR-pins-SerialPort-in-Linux
and https://github.com/PlusToolkit/ndicapi/blob/master/ndicapi_serial_unix.cxx#L162

Not sure about Apple, but it would be great to include all OS's.
In case it helps: https://github.com/PlusToolkit/ndicapi/blob/master/ndicapi_serial_apple.cxx

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
```diff
@@ -562,12 +562,12 @@ PlusStatus vtkPlusUSDigitalEncodersTracker::ReadConfiguration(vtkXMLDataElement*
     encodertrackingInfo.Encoder_PortName = portName;
 
     // ---- Get a name of transformation
-    const char* fromAttribute = EncoderInfoElement->GetAttribute("From");
-    const char* toAttribute = EncoderInfoElement->GetAttribute("To");
+    const char* fromAttribute = EncoderInfoElement->GetAttribute("Id");
```
adamrankin: Local variable should be lower camel case.

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -56,6 +56,8 @@ vtkPlusGenericSerialDevice::vtkPlusGenericSerialDevice()
   : Serial(new SerialLine())
   , SerialPort(1)
   , BaudRate(9600)
+  , DTR(false)
```
adamrankin: Could the variable be more descriptive? Does this mean the requested value of the DTR/DTS pin? If so, something like "RequestedDTR" or "RequestedDTRPinState"

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
```diff
@@ -246,8 +307,20 @@ PlusStatus vtkPlusGenericSerialDevice::ReceiveResponse(std::string& textReceived
...
       {
         // waiting time expired
-        LOG_ERROR("Failed to read complete line from serial device. Received: " << textReceived);
-        return PLUS_FAIL;
+        if (acceptReply == REQUIRE_LINE_ENDING)
+        {
+          LOG_ERROR("Failed to get a proper response within configured time (" << this->MaximumReplyDurationSec << " sec)");
+          return PLUS_FAIL;
+        }
+        else if (textReceived.empty() && acceptReply == REQUIRE_NOT_EMPTY)
```
adamrankin: Switch the order to check REQUIRE_NOT_EMPTY first?

src/PlusDataCollection/vtkPlusGenericSerialDevice.h
```diff
@@ -60,19 +60,40 @@ class vtkPlusDataCollectionExport vtkPlusGenericSerialDevice : public vtkPlusDev
...
+  PlusStatus SetDTR(bool onOff);
+
+  /*! Gets the DTR (data-terminal-ready) line. */
+  vtkGetMacro(DTR, bool);
+
+  /*! Sets the RTS (request-to-send) line. */
+  PlusStatus SetRTS(bool onOff);
+
+  /*! Gets the RTS (request-to-send) line. */
+  vtkGetMacro(RTS, bool);
```
adamrankin: This doesn't actually query the serial device though, does it? The comment should reflect the cached nature of the variable

src/Documentation/UserManual/DeviceGenericSerial.dox
dzenanz: Just in the documentation?

src/PlusDataCollection/PlusSerialLine.cxx
dzenanz: Just adding support for RTS and DTR pins wouldn't be hard, but non-windows part of the SerialLine is totally unimplemented. And I don't have a serial device which I could use for testing. After writing this code I had to ask a collaborator to check whether it is working correctly.

src/Documentation/UserManual/DeviceGenericSerial.dox
adamrankin: Well, hopefully within classes as well. Anything that occurs in an XML file.

dzenanz: This is only a parameter to a method, and the other parameter to the same method is called `textReceived`. Hence I used the same style.

adamrankin: Oh, yes, I see. Why is a method argument in the user manual though?

src/PlusDataCollection/vtkPlusGenericSerialDevice.h
```diff
@@ -60,19 +60,40 @@ class vtkPlusDataCollectionExport vtkPlusGenericSerialDevice : public vtkPlusDev
...
   */
-  virtual PlusStatus SendText(const std::string& textToSend, std::string* textReceived = NULL);
+  virtual PlusStatus SendText(const std::string& textToSend, std::string* textReceived = NULL,
+    ReplyTermination acceptReply = REQUIRE_LINE_ENDING);
 
   /*!
     Receive a response from the serial device.
   */
-  virtual PlusStatus ReceiveResponse(std::string& textReceived);
+  virtual PlusStatus ReceiveResponse(std::string& textReceived, ReplyTermination acceptReply = REQUIRE_LINE_ENDING);
```
adamrankin: Ah, I see Andras' review now. Yes, this only exposes the behavior option to developers. Please make this a parameter modifiable from the XML (see other examples in ReadConfiguration).

src/Documentation/UserManual/DeviceGenericSerial.dox
dzenanz: Maybe I misunderstood Andras' sentence "Please make the AcceptReply criteria configurable and update DeviceGenericSerial.dox"? Or maybe he misunderstood that for a [device](http://www.velmex.com/Downloads/Spec_Sheets/VXM%20-%20%20Command%20Summary%20Rev%20B%20814.pdf) we are dealing with, both settings must be supported within a same session - replies to some commands are CR terminated while a replies to a few commands are not terminated.

adamrankin: Yes, he meant configurable from XML (by the user).

I see. Who is driving the generic serial device? The USEncoder device?

src/PlusDataCollection/vtkPlusGenericSerialDevice.h
dzenanz: This needs to be specified per command. See my other comment.

src/Documentation/UserManual/DeviceGenericSerial.dox
dzenanz: The Velmex VXM motors are the device which needs this. And the exact movement of these motors is being tracked by US Digital encoders (a different device).

adamrankin: Ok, is there also going to be a device class for those motors?

Now that I understand, a method argument is fine. Perhaps a small comment as to which device class needs the behavior, and that different replies have different format for that device.

src/PlusDataCollection/vtkPlusGenericSerialDevice.h
dzenanz: RTS and DTR lines are not used for querying - CTS (clear to send) and DSR (Data Set Ready) are. See [this](https://stackoverflow.com/questions/957337/what-is-the-difference-between-dtr-dsr-and-rts-cts-flow-control). Since only our side can change RTS and DTR it is OK for their values to be cached.

adamrankin: Cool :+1:

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
dzenanz: If this is not descriptive enough, I guess CurrentDTR is the best name for it. But that would screw up `vtkGetMacro(RTS, bool);`. Do you still want me to change it?

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
dzenanz: Done

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
dzenanz: Done

src/PlusDataCollection/vtkPlusGenericSerialDevice.cxx
adamrankin: No, leave it.

src/PlusDataCollection/PlusSerialLine.cxx
adamrankin: Ah yes I see. I may try to refactor to use the ndicapi interface, as it does all of the serial business for us.

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
```diff
@@ -562,12 +562,12 @@ PlusStatus vtkPlusUSDigitalEncodersTracker::ReadConfiguration(vtkXMLDataElement*
...
     // ---- Get a name of transformation
-    const char* fromAttribute = EncoderInfoElement->GetAttribute("From");
-    const char* toAttribute = EncoderInfoElement->GetAttribute("To");
+    const char* fromAttribute = encoderInfoElement->GetAttribute("Id");
+    const char* toAttribute = deviceConfig->GetAttribute("ToolReferenceFrame");
 
     if (!fromAttribute || !toAttribute)
     {
-      LOG_ERROR("Failed to read transform of CoordinateDefinitions (nested element index: " << EncoderInfoElement << ") - check 'From' and 'To' attributes in the configuration file!");
+      LOG_ERROR("Device's attribute 'ToolReferenceFrame' and DataSource's attribute 'Id' are required!");
```
adamrankin: Could it not be reasonable to fallback to the device ID if the ToolReferenceFrame is missing? I think we do this in another/other devices.

dzenanz: Should we fall back to `"Reference"` or `deviceConfig->GetAttribute("Id")`? The second one is strange to me.

adamrankin: There may be multiple trackers in a channel, so if both weren't set, you'd get CameraToReference and CameraToReference (example). With the second option, you'd get CameraToDevice1 and CameraToDevice2

dzenanz:
@lassoan @adamrankin More requests or merge?

dzenanz:
Thanks!

PR #68, opened 2017-08-09
re #1061 Supporting multiple coreXY configurations

dzenanz:
internal class cleanup, removing serial numbers, simplifying the class, reducing duplication, removing persistence and PreTMatrix, adding per DataSource ToolReferenceFrame in order to support compounding transforms by physically combining stages

### DIFF
--- a/src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
+++ b/src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
@@ -185,7 +185,21 @@ PlusStatus vtkPlusUSDigitalEncodersTracker::InternalConnect()
 PlusStatus vtkPlusUSDigitalEncodersTracker::InternalDisconnect()
 {
   LOG_TRACE("vtkPlusUSDigitalEncodersTracker::Disconnect");
-  return this->StopRecording();
+  this->StopRecording();
+
+  if (::IsInitialized() != 1)
+  {
+      // Device not yet initialized
+      return PLUS_SUCCESS;
+  }
+
+  if (::CloseSEI() != 0)
+  {
+      LOG_ERROR("Failed to close SEI!");
+      return PLUS_FAIL;
+  }
+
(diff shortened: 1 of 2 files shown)


src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
```diff
@@ -364,148 +252,94 @@ PlusStatus vtkPlusUSDigitalEncodersTracker::InternalUpdate()
...
       }
     }
-  }
 
-  for (unsigned int i = 0; i < USDigitalEncoderTrackingInfoList.size(); ++i)
-  {
-    if (!USDigitalEncoderTrackingInfoList[i].Encoder_Persistent)
+    it->TransformationMatrix->DeepCopy(tempTransform->GetMatrix());
+    this->TransformRepository->SetTransform(it->TransformName, it->TransformationMatrix);
+    if (MyToolTimeStampedUpdate(*it) == PLUS_FAIL)
```
adamrankin: I admit I don't love the "My" nomenclature as it is not descriptive. I am open to a discussion though as I see it is cleaner. Something in between?

adamrankin:
One thing I see is that InternalStopRecording performs a full disconnect (CloseSEI) but InternalStartRecording does not perform a reconnect.

If the user send a command to stop and start recording, this device will not resume collecting data (unless A2GetPosition or similar) will automatically resume after CloseSEI?

Should CloseSEI be in InternalDisconnect and not InternalStopRecording?

src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
dzenanz: Since it is invoked in just one place, it can be completely eliminated.

dzenanz:
You are right about StopRecording/Disconnect. Fix coming.

dzenanz:
@jperdomo23 @jamesobutler Can you check that the last commit does not break anything?

PR #72, opened 2017-08-22
Scoliosis transverse process enhancer

travers-bryan:
Continued and finalized the implemented of the vtkPlusTransverseProcessEnhancer.cxx algorithm. This program allows for real time interpretation of ultrasound scans of a patient with scoliosis.

The header file vtkPlusTransverseProcessEnhancer.h was also implemented appropriately.
The file vtkPlusTransverseProcessEnhancerTest.cxx has also been updated to allow for better testing of these algorithms.


src/PlusImageProcessing/Testing/vtkPlusTransverseProcessEnhancerTest.cxx
```diff
@@ -6,7 +6,7 @@ See License.txt for details.
 
 /*!
 \file vtkPlusTransverseProcessEnhancerTest.cxx
-\brief TODO Give a one-liner explanation as to the basics of this test
+This is a program meant to test vtkPlusTransverseProcessEnhancer.cxx from the command line.
```
adamrankin: Perhaps be a bit more specific like "This program loads an XYZ sequence from file and generates some ABC data..." or something

src/PlusImageProcessing/Testing/vtkPlusTransverseProcessEnhancerTest.cxx
```diff
@@ -16,30 +16,205 @@ See License.txt for details.
...
+  vtkSmartPointer<vtkPlusTransverseProcessEnhancer> enhancer;
+  enhancer = vtkSmartPointer<vtkPlusTransverseProcessEnhancer>::New();
+    
+  // Read input sequence
+  vtkSmartPointer<vtkPlusTrackedFrameList> trackedFrameList = vtkSmartPointer<vtkPlusTrackedFrameList>::New();
+  if (trackedFrameList->ReadFromSequenceMetafile(inputFileName) == PLUS_FAIL)
+  {
+    return EXIT_FAILURE;
+  }
+  vtkSmartPointer<vtkImageCast> castToUchar = vtkSmartPointer<vtkImageCast>::New();
```
adamrankin: Remove if no longer used

src/PlusImageProcessing/vtkPlusTransverseProcessEnhancer.h
```diff
@@ -41,189 +45,147 @@ class vtkPlusImageProcessingExport vtkPlusTransverseProcessEnhancer : public vtk
...
-  /*! Set optional output file name for sub-sampled input image sequence */
-  void SetLinesImageFileName(const std::string& fileName);
-  
-  /*! Set optional output file name for transverse process acoustic shadow images */
-  void SetShadowImageFileName(const std::string& fileName);
-
-  /*! Set optional output file name for partially process images */
-  void SetIntermediateImageFileName(const std::string& fileName);
+  /*! If optional output files for intermediate images should saved */
+  vtkSetMacro(IntermediateImageFileName, std::string);
```
adamrankin: Please use the vtkSetStdStringMacro

src/PlusImageProcessing/vtkPlusTransverseProcessEnhancer.h
```diff
@@ -4,12 +4,15 @@ Copyright (c) Laboratory for Percutaneous Surgery. All rights reserved.
...
 
+
+
 #ifndef __vtkPlusTransverseProcessEnhancer_h
 #define __vtkPlusTransverseProcessEnhancer_h
 
 // Local includes
 #include "vtkPlusImageProcessingExport.h"
 #include "vtkPlusTrackedFrameProcessor.h"
+#include "vtkImageAlgorithm.h"
```
adamrankin: I may have missed it, but it seems like this isn't necessary

src/PlusImageProcessing/vtkPlusTransverseProcessEnhancer.cxx
```diff
@@ -24,125 +24,100 @@ See License.txt for details.
...
+  LinesImage(NULL),
+  ProcessedLinesImage(NULL),
+  UnprocessedLinesImage(NULL)
+
 {
-  this->SetDilationKernelSize(0, 0);
+  this->SetMmToPixelFanImage(0, 0, 0);
+  this->SetMmToPixelLinesImage(0, 0, 0);
+
+  this->GaussianSmooth = vtkSmartPointer<vtkImageGaussianSmooth>::New();    //Used to smooth the image
```
adamrankin: Please return these to the constructor list

adamrankin: Yes, see, here. I see, he's getting two different instructions.

So, my original question, is this a style or performance reason?

cpinter: Style, but I guess for smart pointers where deleting them is not a concern they are fine in the constructor list.

adamrankin: Ok, I don't mind having them in the body. I'll switch.

lassoan: In recent nightly versions, vtkNew<> does not require .GetPointer() anymore to get the pointer, so we can use that most places instead of vtkSmartPointer<>.
vtkNew<> works in the header, so no additional code will be needed in the constructor.

cpinter:
I can change the smart pointers for vtkNew. But because Bryan left, it would be very hard to amend this pull request, as it is from his fork. I suggest that we integrate this, then I'll do whatever changes necessary.

PR #73, opened 2017-09-07
re #1194 Supporting focal depth setting. Requires WinProbeSDK-2.0

dzenanz:


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -69,6 +72,15 @@ PlusStatus vtkPlusWinProbeVideoSource::ReadConfiguration(vtkXMLDataElement* root
         }
     }
     
+    const char * fpdCharPtr = deviceConfig->GetAttribute("FocalPointDepth");
```
adamrankin: Consider using the Plus xml macros. If one doesn't exist for a vector 4. feel free to create it.

lassoan: There is standard VTK method for this:  vtkXMLDataElement::GetVectorAttribute
https://www.vtk.org/doc/nightly/html/classvtkXMLDataElement.html#a3bd5f80f4cab1fde643409eca4bb7ea7

PR #74, opened 2017-09-11
re #1194 Simplifying code by using existing XML vector methods

dzenanz:
This is a follow-up to #73. I tested it, the new code behaves the same.

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
@@ -134,8 +134,8 @@ protected:
     PlusTrackedFrame::FieldMapType m_customFields;
     std::thread * m_watchdog32 = nullptr;
     double m_lastTimestamp = 0.0; //for watchdog
-    double m_timeGainCompensation[8];
-    float m_focalPointDepth[4];
+    double m_timeGainCompensation[8] = { 0 };
+    float m_focalPointDepth[4] = { 0 };
 
 private:
     vtkPlusWinProbeVideoSource(const vtkPlusWinProbeVideoSource &); // Not implemented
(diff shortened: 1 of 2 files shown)


dzenanz:
@cpinter @lassoan @adamrankin Review and merge?

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
```diff
@@ -134,8 +134,8 @@ class vtkPlusDataCollectionExport vtkPlusWinProbeVideoSource : public vtkPlusUsD
     PlusTrackedFrame::FieldMapType m_customFields;
     std::thread * m_watchdog32 = nullptr;
     double m_lastTimestamp = 0.0; //for watchdog
-    double m_timeGainCompensation[8];
-    float m_focalPointDepth[4];
+    double m_timeGainCompensation[8] = { 0 };
```
adamrankin: Does this compile on VS2010?

@lassoan Do we care anymore about VS2010? Can we say that VS2010 is for stealthlink only?

dzenanz: I kind of remember a recent declaration that Plus is now C++11 project, because VTK is too.

adamrankin: Ok, fine by me then.

lassoan: It should be fine, we use this kind of initialization at many places in the code. Not related to C++11.

Note that VTK is very far from being a C++11 project, but a few C++11 features are allowed to be used in VTK now (see https://docs.google.com/document/d/1nzinw-dR5JQRNi_gb8qwLL5PnkGMK2FETlQGLr10tZw/edit#heading=h.gc38h63576nk). Of course those features are allowed in Plus, too, but in Plus we may decide to be a bit more permitting, if needed.

dzenanz: Oops, this does not compile in VS2013. I suppose that is a problem?

lassoan: Oh, sorry, it's in the header - it does not work there.

lassoan: Just move initialization to the constructor

PR #75, opened 2017-09-12
re #1149 New version of BK scanner interface

olevs:
Initial version of the new BK scanner interface. This version uses vtkClientSocket to replace GrabbieLib. It is tested with the BK5000 scanner. Currently the code works on OS X and Linux, but not on Windows. The immediate problem on Windows is that vtkClientSocket fails to create a socket.


olevs:
I can also make a new version of PlusDeviceSet_Server_BkProFocusOem.xml and put this in PlusLibData. Can this be placed in https://github.com/PlusToolkit/PlusLibData?

adamrankin:
Does it make sense to remove the GrabbieLib implementation? Should this not be a separate BkNetwork video source?

Or is this network interface backwards compatible with all the older scanners?

Also, we will wait for the merge until the windows socket issue is sorted. Do you know what may be causing the problem?

lassoan:
> I can also make a new version of PlusDeviceSet_Server_BkProFocusOem.xml and put this in PlusLibData. Can this be placed in https://github.com/PlusToolkit/PlusLibData, or are the move process not finished yet?

The moving of repositories is completed, so you can just send a pull request with the added/updated file.

lassoan:
Socket works well for me on Windows10. I've tested with this Python server:

<pre>
import socket

TCP_IP = '127.0.0.1'
TCP_PORT = 5005
BUFFER_SIZE = 20  # Normally 1024, but we want fast response

s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.bind((TCP_IP, TCP_PORT))
s.listen(1)

conn, addr = s.accept()
print 'Connection address:', addr
while 1:
    data = conn.recv(BUFFER_SIZE)
    print "received connection"
    if not data: break
    print "received data:", data
    conn.send(data)  # echo
conn.close()
</pre>

It prints this when I connect using vtkBkProFocusOemVideoSourceTest:

<pre>
waiting for connection
received connection
Connection address: ('127.0.0.1', 57589)
received data: QUERY:US_WIN_SIZE;
received data: QUERY:B_GEOMETRY_SC
received data: ANAREA:A;QUERY:B_G
received data: EOMETRY_PIXEL:A;QU
received data: ERY:B_GEOMETRY_US_FR
received data: AME_GRAB:A;QUERY:B
received data: _GEOMETRY_TISSUE:A;
received data: QUERY:B_GAIN:A;QU
received data: ERY:TRANSDUCER_LIST;
received data: QUERY:TRANSDUCER:A
received data: ;CONFIG:EVENTS 1;
received data: CONFIG:DATA:SUBSCRI
received data: BE "US_WIN_SIZE","B_
received data: GEOMETRY_SCANAREA","
received data: B_GEOMETRY_PIXEL","B
received data: _GEOMETRY_US_FRAME_G
received data: RAB","B_GEOMETRY_TIS
received data: SUE","B_GAIN","TRANS
received data: DUCER";
received data: query:capture_image
received data:  "PNG";
</pre>

Can you provide instructions for reproducing your socket problem?

olevs:
@lassoan Great that it works for you. Then I assume the problem is with my rather old computer running Windows 10. On my Windows machine both PlusServer and vtkBkProFocusOemVideoSourceTest either fails or crashes when trying to use vtkClientSocket.

@adamrankin This interface should in theory be backwards compatible with the other scanners, as it  uses the same OEM interface as GrabbieLib, so the best solution is to replace it. In practice there may of course be issues. For example: I have observed some strange behavior from the BK5000 scanner where it seems that is don't work according to the specification. The problems I have found are handled in the current code, and I also just sent a mail to BK about it. There may still be problems I haven't found, or different issues with different scanners.

About removing GrabbieLib completely. I have not done any work with vtkPlusBkProFocusCameraLinkVideoSource and PlusBkProFocusCameraLinkReceiver. These also must be rewritten if you want to remove all dependencies of GrabbieLib.

adamrankin:
I think unless the replacement of GrabbieLib is complete, it should stay in PlusLib.

And if the change isn't backwards compatible (in practice) then I wouldn't recommend replacing an existing device type, but add a new one.

I will let @lassoan have the final say.

olevs:
Both solutions are fine with me. Either replacing the existing device or creating a new one. If we are to have two devices we may want to have two different config files. The new xml config file I made for the new version of the BK scanner interface will not work with the old scanner interface without modifications. 

@lassoan What do you think about replacing the existing device or making a duplicate?

Feel free to modify my fork if there are changes you want to make.

If we are to make a duplicate device the build system also have to be changed: https://github.com/PlusToolkit/PlusBuild/pull/9

lassoan:
@olevs Sorry for the long delay. I've reviewed the changes and they all look good. There is no need to keep the old BK OEM interface class - the old BK CameraLink interface class is independent from that. Could you please rebase the branch based on the latest master (there are some merge conflicts)? If not then let me know and I'll try to resolve the conflicts. Thank you!

olevs:
Great. I merged in the latest changes from the master branches of PlusLib and PlusBuild.

lassoan:
Thank you!

PR #76, opened 2017-09-13
re #1194 Fixing initialization bug and VS2013 compile error

dzenanz:
Due to initialization bug in constructor, I thought these arrays were not initialized - hence fiddling with it in the previous commit.


dzenanz:
@adamrankin Review this too?

PR #77, opened 2017-09-14
Use device IDs instead of device addresses

dzenanz:
This handles the case of addresses assigned not starting at 0, which happens when connecting and then disconnecting additional encoders.


src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
```diff
@@ -134,6 +134,7 @@ PlusStatus vtkPlusUSDigitalEncodersTracker::InternalConnect()
     LOG_ERROR("Failed to initialize SEI!");
     return PLUS_FAIL;
   }
+  IdAddress.clear();
```
adamrankin: Minor coding standard

member access with this->

PR #175, opened 2017-09-21
Adding support for OptiTrack running in background or GUI mode

markasselin:
Added support for using OptiTrack devices with Motive running in the background (via Motive API) or with the Motive GUI running (via NatNet SDK).


lassoan:
👍

PR #182, opened 2017-09-26
Supporting WinProbeSDK 2.1

dzenanz:
Some functions got renamed, one of them was being called from Plus wrapper. @jamesobutler can you test whether this is still working correctly with the latest SDK?


jamesobutler:
ok, I'll let you know if it works.

RMR54:
@dzenanz, just tested with latest Winprobe SDK and it works. Thanks

Issue #160, opened 2017-09-21
Latest PLUS cannot connect to Interson probe


Sunderlandkyl:
I am using the Interson USB ultrasound probe (GV 5.0). I cannot connect to the the probe using the latest. Both the stable release PlusApp-2.4.0.5112-Interson-MTC-3.7-Win32 and the nightly PlusApp-2.5.0.20170722-Interson-MTC-3.7-Win32 exhibit this behaviour.

Note that I can successfully connect using an older version of PLUS: PlusApp-2.2.0.4177-Interson-MTC-3.6-Win32.

I am using the standard configuration file that is distributed with the download (I've attached it for convenience).PlusDeviceSet_Server_IntersonVideoCapture.xml

Error log:
~~~~
|INFO|000.042000| Software version: Plus-2.5.0.dbb84f3 - Win32| in D:\D\PTNPIMT37b\PlusApp\PlusServerLauncher\PlusServerLauncherMainWindow.cxx(106)
|INFO|000.042000| Logging at level 3 to file: C:/Users/mholden8/PlusApp-2.5.0.20170722-Interson-MTC-3.7-Win32/data/080817_150945_PlusLog.txt| in D:\D\PTNPIMT37b\PlusApp\PlusServerLauncher\PlusServerLauncherMainWindow.cxx(107)
|INFO|000.045000| Supported devices: - 3dConnexion (ver: Plus-2.5.0) - Ascension3DG (ver: Plus-2.5.0) - AuroraTracker (ver: NDICAPI-1.7) - BrachyTracker (ver: Plus-2.5.0) - CertusTracker (ver: Optotrak API (Win32 DLL Version) Version 3.01.03) - ChRobotics (ver: Plus-2.5.0) - Epiphan (ver: Plus-2.5.0) - FakeTracker (ver: Plus-2.5.0) - GenericSerialDevice (ver: Plus-2.5.0) - ICCapturing (ver: The Imaging Source UDSHL-3.4) - ImageProcessor (ver: Plus-2.5.0) - IntersonVideo (ver: Interson Bmode DLL v243, USB Probe DLL v163| in D:\D\PTNPIMT37b\PlusApp\PlusServerLauncher\PlusServerLauncherMainWindow.cxx(111)
|INFO|000.047000| Server host name: aorta| in D:\D\PTNPIMT37b\PlusApp\PlusServerLauncher\PlusServerLauncherMainWindow.cxx(131)
[...]

PR #187, opened 2017-09-29
re #160: Fix issue with Interson probe connection

Sunderlandkyl:


src/PlusDataCollection/Interson/vtkPlusIntersonVideoSource.cxx
```diff
@@ -122,7 +122,7 @@ class vtkPlusIntersonVideoSource::vtkInternal
   }
 
   //----------------------------------------------------------------------------
-  PlusStatus vtkPlusIntersonVideoSource::vtkInternal::InitializeDIB(int imageSize[2])
+  PlusStatus vtkPlusIntersonVideoSource::vtkInternal::InitializeDIB(std::vector<int> imageSize)
```
adamrankin: @lassoan Are we still coding const & or are we letting the compiler decide?

lassoan: const& would be more appropriate

We should also replace std::vector<int> with a typedef and the std::vector<int> with a fixed-size class. We might use either itkFixedArray or just redefine it in Plus to avoid ITK dependency in such a core type.

adamrankin: I am very happy with std::array<type, size> as it is iterator compatible.

lassoan: Agreed. Such change may cause regressions, so we should do it after we released the new stable version (in a few weeks). I've added a ticket to track that task (#188).

Sunderlandkyl: Alright, I've added const & to the function definition.

lassoan:
Thank you looks good to me. Before we merge, could you please add these to the commit comment:

- Fixed image artifacts at the image boundary due to too small image window.
- Fixed crash during connection due to uninitialized image size (regression introduced during image size type refactoring many months ago)
- Fixed storage of custom fields in image buffer.

If it's not too much trouble, it would be nice to split the vtkPlusUsDevice::AddVideoItemToVideoSources modification into a separate commit.

adamrankin:
Crap, sorry.

lassoan:
It's OK, I was too slow.

Sunderlandkyl:
I can still amend the commit to fix the comment and separate the fixes if you want.

lassoan:
It would be nice, thank you (if you do it, then you have to do it quickly to avoid causing merge conflicts for people who pull the current version).

Issue #196, opened 2017-10-03
Support multiple tracking tools in StealthLink


lassoan:
There are some limitations in what tools can be tracked simultaneously by StealthStation, but all tools should be accessible through StealthLink:

lassoan:
You can use TESLA to build a Plus package with StealthLink support. The build tree is here: d:\D\PTNPSLb

lassoan:
from @Sunderlandkyl:

> I made some modifications to Plus so that it could track all of the tools in the FOV, instead of just 1.
> It works, however the execution time of InternalUpdate has gone from ~0.043s to ~0.188s (for 3 tools and 1 reference frame).
> Specifically, querying the server for instrument data takes about ~0.043s per tracked object.
> 
> I'm going to keep looking at the documentation to see if I can find an alternative method.

lassoan:
Thanks. It sounds promising, however the dropping of the frame rate from 23fps to 5fps may not be acceptable in many applications.

@ungi Do we need to track more tools than what StealthStation supports?

If yes, then we'll have to contact Medtronic to see if really 5fps is the best performance we can get. If not, then just commit what we have (but make sure the performance is not degraded compared to what we had before).

Sunderlandkyl:
@lassoan If I make the optimization that we discussed today (removing a redundant query). The fps would be something like:

- 1 frame + 1 tool: 23.3fps
- 1 frame + 2 tools: 11.6fps
- 1 frame + 3 tools: 7.8fps
- 1 frame + 4 tools: 5.8fps

This should give us the same performance as before when using 1 frame and 1 tool, but would still decrease in performance as more tools are added.

Perhaps we could add an option in the config file to either track all tools, or to match the StealthStation's behavior and only track the tool that is closest to the frame?

lassoan:
> Perhaps we could add an option in the config file to either track all tools

Good idea. Let's add an option for that. Probably the best would be to add it in the DataSource element (AlwaysTrack? ForceQuery? or something like that) so that it can be set for specific tools only, by default disabled (better performance).

PR #197, opened 2017-10-04
re #196: Add support for multiple tracked tools in StealthLink

Sunderlandkyl:
By adding the AlwaysTrack="TRUE" attribute to the DataSource, the tool will always be tracked, even if it is not the closest tool to the reference frame. 

`<DataSource Type="Tool" Id="" PortName="" AlwaysTrack="TRUE" />`

If not specified, AlwaysTrack is assumed to be false.

Relevant ticket: https://github.com/PlusToolkit/PlusLib/issues/196

### DIFF
--- a/src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
+++ b/src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
@@ -88,18 +88,32 @@ public:
     return this->ExamValid;
   }
 
+  //----------------------------------------------------------------------------
   // Get for Navigation Data. It is called each time internalupdate is called
   PlusStatus GetCurrentNavigationData(MNavStealthLink::NavData& navData)
   {
     MNavStealthLink::Error err;
+    PlusLockGuard<vtkPlusRecursiveCriticalSection> updateMutexGuardedLock(this->StealthLinkServerMutex);
+    MNavStealthLink::DateTime myDateTime = this->StealthLinkServer->getServerTime();
+    if (!this->StealthLinkServer->get(navData, myDateTime, err))
     {
-      PlusLockGuard<vtkPlusRecursiveCriticalSection> updateMutexGuardedLock(this->StealthLinkServerMutex);
-      MNavStealthLink::DateTime myDateTime = this->StealthLinkServer->getServerTime();
-      if (!this->StealthLinkServer->get(navData, myDateTime, err))
-      {
-        LOG_ERROR(" Failed to acquire the navigation data from StealthLink Server: " <<  err.reason() << " " << err.what() << "\n");
-        return PLUS_FAIL;
-      }
+      LOG_ERROR(" Failed to acquire the navigation data from StealthLink Server: " <<  err.reason() << " " << err.what() << "\n");
+      return PLUS_FAIL;
+    }
+    return PLUS_SUCCESS;
+  }
+
+  //----------------------------------------------------------------------------
+  // Get for Instrument Data. It is called each time internalupdate is called for any additional instruments that need to be tracked
+  PlusStatus GetInstrumentData(MNavStealthLink::Instrument& instrument, const std::string toolname)
+  {
+    MNavStealthLink::Error err;
+    PlusLockGuard<vtkPlusRecursiveCriticalSection> updateMutexGuardedLock(this->StealthLinkServerMutex);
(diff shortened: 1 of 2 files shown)


src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -105,6 +106,23 @@ class vtkPlusStealthLinkTracker::vtkInternalShared
   }
 
   //----------------------------------------------------------------------------
+  // Get for Instrument Data. It is called each time internalupdate is called for any additional instruments that need to be tracked
+  PlusStatus GetInstrumentData(MNavStealthLink::Instrument& instrument, const std::string toolname)
+  {
+    MNavStealthLink::Error err;
```
lassoan: why this variable is defined outside the {} block?

src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -917,10 +987,36 @@ PlusStatus vtkPlusStealthLinkTracker::InternalUpdate()
         this->ToolTimeStampedUpdateWithoutFiltering(toolIterator->second->GetSourceId(), instrumentToTrackerTransform, TOOL_OUT_OF_VIEW, unfilteredTime, timeSystemSec);
       }
     }
+    // if the wanted Transform is any tool to Tracker that is not considered the current instrument
     else
     {
       vtkSmartPointer<vtkMatrix4x4> transformMatrixForNotTrackedTool = vtkSmartPointer<vtkMatrix4x4>::New();
-      this->ToolTimeStampedUpdateWithoutFiltering(toolIterator->second->GetSourceId(), transformMatrixForNotTrackedTool, TOOL_OUT_OF_VIEW, unfilteredTime, timeSystemSec);
+      // if we are tracking the current instrument at all times
```
lassoan: Write here a bit more about what is the current instrument (closest to the reference?), why we don't track all instruments by default (due to time it takes for the query to compete), ...

src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -917,10 +987,36 @@ PlusStatus vtkPlusStealthLinkTracker::InternalUpdate()
...
     {
       vtkSmartPointer<vtkMatrix4x4> transformMatrixForNotTrackedTool = vtkSmartPointer<vtkMatrix4x4>::New();
-      this->ToolTimeStampedUpdateWithoutFiltering(toolIterator->second->GetSourceId(), transformMatrixForNotTrackedTool, TOOL_OUT_OF_VIEW, unfilteredTime, timeSystemSec);
+      // if we are tracking the current instrument at all times
+      if (STRCASECMP(toolIterator->second->GetCustomProperty("AlwaysTrack").c_str(), "TRUE") == 0)
+      {
+        MNavStealthLink::Instrument instrument;
+        if (!this->InternalShared->GetInstrumentData(instrument, toolIterator->second->GetPortName()))    //thread safe
+        {
+          return PLUS_FAIL;
```
lassoan: Does GetInstrumentData return bool or PlusStatus? In general, use PlusStatus to return success/fail.
Does GetInstrumentData log the error? If not, then log the error before returning.

src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -917,10 +987,36 @@ PlusStatus vtkPlusStealthLinkTracker::InternalUpdate()
...
+        }
+
+        vtkPlusStealthLinkTracker::vtkInternal::GetInstrumentInformation(instrument, instrumentOutOfView, instrumentToTrackerTransform);
+        if (instrumentOutOfView == false)
```
lassoan: I know that instrumentOutOfView has been in the code, but It would make the code easier to read if we didn't use negative terms. It would be more clear to write "instrumentInTrackerView" than "instrumentOutOfView == false". We don't have to change it now, but we should consider. Or, instead of a binary flag, we could use TOOL_OK, TOOL_OUT_OF_VIEW enum.

Sunderlandkyl:
@lassoan I've made some changes, how does it look now?

Sunderlandkyl:
@lassoan Do we want to merge these changes before or after the new stable?

PR #203, opened 2017-10-06
Enabling doxygen xml output in anticipation of markdown doxygen output

adamrankin:


adamrankin:
@lassoan Could you just confirm that renaming generate_doc.cmake to GenerateDoc.cmake won't have any effects elsewhere?

Issue #205, opened 2017-10-11
Swap R and B channels in the device OpenCV_VIDEO


carlos-luque:
The device OpenCV_Video does not show the RGB codification in frames

lassoan:
Has this been resolved?

carlos-luque:
I think it has been partly resolved. When I use the updated code for a device openCV_video, the color codification is correctly shown in the 3D Slicer but is not in the fCal. Both applications use the same configuration file.   I do not know _why_ the application fCal does not take the updated color codification.

lassoan:
> color codification is correctly shown in the 3D Slicer but is not in the fCal

Do you mean that R and B color channels are swapped when you look at the image in fCal?

carlos-luque:
Yes, I do.

lassoan:
OK. We'll investigate this after the new stable version is released.

PR #206, opened 2017-10-11
re #205: exchanging between BGR codification and RGB codification

carlos-luque:
I have fixed the frame RGB codification in the function vtkPlusOpenCVCaptureVideoSource::InternalUpdate().


lassoan:
Thanks for the pull request.

@markasselin Could you test this to see if color appears as we expect and there is no significant performance impact? Thanks!

adamrankin:
Just fyi that this is a GPU optimized function, and so performance will be dependent on the hardware available.

lassoan:
If image data is in CPU memory then RGB<->BGR conversion is probably slower in GPU than using CPU MMX instructions. Hopefully, the OpenCV implementation is smart enough to choose GPU/CPU based on what is more efficient to use.

adamrankin:
Confirmed with our student who worked with the NDI Vega RTSP stream that indeed the colours were reversed, as so this fixes that problem!

Thanks Carlos!

PR #239, opened 2017-10-19
Use hardware frame numbering for Capistrano

dzenanz:


dzenanz:
@jamesobutler @jperdomo23 Can somebody check that this PR does not cause problems?

PR #244, opened 2017-10-25
Fix for connection issues with NDI Polaris

Sunderlandkyl:
Fixed or Irrelevant Issues:
  - Connecting with a baud rate of 19200 always fails with Device->host communication timeout [(Line 973](https://github.com/Sunderlandkyl/PlusLib/blob/e6bc96cd4038613975df002ef27afdb82cd7bd19/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx#L973)) - probably due to this baud rate is reused for indicating higher speed => document that we do not recommend this baud rate
  - Connecting with a baud rate of 14400 results in a "Command is invalid in current mode" error => document that we do not recommend this baud rate
  - Frequent errors caused by the Probe() function.
    - Fixed in https://github.com/PlusToolkit/ndicapi/pull/7
  - Frame number is incorrect: => test if it is a parsing issue
    - The frame number returned at [Line 542](https://github.com/Sunderlandkyl/PlusLib/blob/e6bc96cd4038613975df002ef27afdb82cd7bd19/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx#L542) switches between a small and large number every couple seconds. (ex 16767 and 4294967170).
    - Fixed in: https://github.com/Sunderlandkyl/ndicapi/commit/bef0e4400f626eae42e43ddadcda62ed436abb80
  - Infrequent failure (1 in 5 times) in vtkPlusNDITracker::EnableToolPorts() with some baud rates: (1228739, 921600, 230400) => investigate a bit more to see if delays or retries can fix it
    - Error returned at [Line 684](https://github.com/Sunderlandkyl/PlusLib/blob/e6bc96cd4038613975df002ef27afdb82cd7bd19/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx#L684)
      -  0x40: Tool definition file error
    - **FIXED**: Errors were occurring during the SROM file upload. The error was only being checked at the end, ignoring potential timeout issues during the upload.
  - Connecting with a baud rate of 230400 fails as Polaris responds with an error (probably the firmware does not support this speed), so Plus tries to automatically find the highest baud rate => no action needed, just make auto-select robust

### DIFF
--- a/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx
+++ b/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx
@@ -1008,7 +1008,7 @@ PlusStatus vtkPlusNDITracker::ReadConfiguration(vtkXMLDataElement* rootConfigEle
 
   XML_READ_SCALAR_ATTRIBUTE_OPTIONAL(unsigned long, SerialPort, deviceConfig);
   XML_READ_SCALAR_ATTRIBUTE_OPTIONAL(unsigned long, BaudRate, deviceConfig);
-  if (vtkPlusNDITracker::ConvertBaudToNDIEnum(this->BaudRate) == -1)
+  if (deviceConfig->GetAttribute("BaudRate") != NULL && vtkPlusNDITracker::ConvertBaudToNDIEnum(this->BaudRate) == -1)
   {
     LOG_WARNING("Invalid baud rate specified, reverting to auto-select.");
     this->BaudRate = 0;
--- a/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx
+++ b/src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx
@@ -84,7 +84,7 @@ vtkPlusNDITracker::vtkPlusNDITracker()
   , Device(nullptr)
   , SerialDevice("")
   , SerialPort(-1)
-  , BaudRate(9600)
+  , BaudRate(0)
   , IsDeviceTracking(0)
   , LeaveDeviceOpenAfterProbe(false)
   , MeasurementVolumeNumber(0)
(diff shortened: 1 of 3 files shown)


adamrankin:
So, reverting the code means the user has to specify the serial port? Other platforms do not have this requirement and it would be better to identify the issues rather than avoid them.

adamrankin:
What compiler are you working with?

Sunderlandkyl:
I'll take another look after I figure out the frame number issue.
Currently I'm using VS2017.

adamrankin:
One thing worth trying is to see if it's the parallelization causing your issues. You can comment out 187, 188 and 189 (in the original, your lines will be different).

Sunderlandkyl:
Sure, I'll give it a try.

adamrankin:
Hmm, I'm using the same thing =/

Sunderlandkyl:
When the probe function is enabled, it fails to detect the device if it is not correctly disconnected and the Polaris is left running at the high baud rate.

adamrankin:
Ok, I will try and replicate that here and fix it! Thanks for all the investigating.

Sunderlandkyl:
I managed to fix the frame number issue and will make a pull request to ndcapi shortly.
Have you managed to replicate the probe issue?

adamrankin:
How do you force it to disconnect improperly? Kill the Plus process?

adamrankin:
I do not experience a reconnection problem after the device is left in a high baud state. The serial break properly resets the device.

Sunderlandkyl:
It happens to me if I just stop debugging.

adamrankin:
I can't replicate that, mine reconnects fine. Are you on Skype? I sent a request to what I think is you.

adamrankin:
Are you guys plugging in the tracker to a control box, and then from the control box to their USB hub thingy, and then to a PC? Or do you go right from camera to their USB hub to PC?

Sunderlandkyl:
I think I've managed to fix it. Just give me a minute to figure out which changes were necessary, and I'll make a pull request at ndicapi.

src/PlusDataCollection/NDICAPITracking/vtkPlusNDITracker.cxx
```diff
@@ -990,6 +1008,11 @@ PlusStatus vtkPlusNDITracker::ReadConfiguration(vtkXMLDataElement* rootConfigEle
 
   XML_READ_SCALAR_ATTRIBUTE_OPTIONAL(unsigned long, SerialPort, deviceConfig);
   XML_READ_SCALAR_ATTRIBUTE_OPTIONAL(unsigned long, BaudRate, deviceConfig);
+  if (vtkPlusNDITracker::ConvertBaudToNDIEnum(this->BaudRate) == -1)
```
adamrankin: I added this:
  if(deviceConfig->GetAttribute("BaudRate") != NULL && vtkPlusNDITracker::ConvertBaudToNDIEnum(this->BaudRate) == -1)
  {

Sunderlandkyl: Do you want me to amend my commit to add that?

adamrankin: Yes please, removes a warning when no baud rate is specified.

Sunderlandkyl:
@adamrankin Thanks for all your help!

PR #255, opened 2017-11-01
Vtk plus transverse process enhancer devel

BenChurch:
Ok, I finally got the rebase to work. It looks like this should be ready to merge. Improves vtkPlusTrasnverseProcessEnhancer bone segmentation with shadow comparison


lassoan:
Yes, this is much better. The only issue is that some of the commits has still conflicts.

image

Could you squash all the commits into one commit?

BenChurch:
Yes, I'll try to figure this out and let you know if there's a problem

PR #257, opened 2017-11-02
Squash vtk plus transverse process enhancer

BenChurch:
I've got the commits squashed but now I get these 3 other modified files. Can they simply be excluded? Otherwise I'll figure out where they come from and keep them out.


Sunderlandkyl:
Merged in 1e82ad62ce29feb84aaad22692a2dde8a9b0b149.
Let me know if you see any problems.

PR #264, opened 2017-11-06
Open Haptics support

sjh26:
This adds support for Open Haptics based devices (https://www.3dsystems.com/haptics-devices/openhaptics) to Plus.  The device reports tracking data for the stylus, along with tooltip velocity and button states.  It can take force inputs as transforms with the force data (3d vector) set as the translation.


src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  //schedule the internal callback - must be done each time device is polled
+  //the scheduler is intended to manage multiple clients requesting data from the same device
+  hdScheduleSynchronous(positionCallback, this, HD_MAX_SCHEDULER_PRIORITY);
+
+  return PLUS_SUCCESS;
+}
+
+PlusStatus vtkPlusOpenHapticsDevice::InternalConnect()
+{
+  if (this->isHapticDeviceInitialized)
```
adamrankin: There is already a parent level variable for this check

see this->Connected

adamrankin: And this is already checked by 
PlusStatus vtkPlusDevice::Connect()

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  {
+    LOG_DEBUG("Already connected - " << this->DeviceName);
+    return PLUS_SUCCESS;
+  }
+
+  if (this->InputChannels.empty())
```
adamrankin: This type of check is better placed in NotifyConfigured, so the system will fail sooner, rather than try to connect to some devices before noticing this failure. Some devices have long connect times, and NotifyConfigured provides an early exit for configuration problems.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+    }
+    vtkPlusDataSource* forceInput;
+    if (!this->InputChannels[0]->GetOwnerDevice()->GetToolByPortName("Force", forceInput))
+    {
+      LOG_WARNING("No Force tool in input channel.  Forces will not be available");
+    }
+  }
+
+  HDErrorInfo errorFlush;
+  while (HD_DEVICE_ERROR(errorFlush = hdGetError())) {}
```
adamrankin: Is this potentially an infinite loop?

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+
+  DeviceHandle = hdInitDevice(this->DeviceName.c_str());
```
adamrankin: Convention is to use this-> for member variables

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+    DeviceHandle = -1;
+    return PLUS_FAIL;
+  }
+
+  // Enable forces
+  hdEnable(HD_FORCE_OUTPUT);
+  hdEnable(HD_FORCE_RAMPING);
+
+  this->isHapticDeviceInitialized = true;
+  LOG_INFO("Phantom initialized: " << this->DeviceName)
```
adamrankin: This may be more appropriate as LOG_DEBUG, however totally up to you, LOG_INFO is not wrong!

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+
+  hdStartScheduler();
+
+  return PLUS_SUCCESS;
+}
+
+PlusStatus vtkPlusOpenHapticsDevice::InternalDisconnect()
+{
+  if (this->isHapticDeviceInitialized)
```
adamrankin: Again, this->Connected and parent level checking already handles this for you.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+PlusStatus vtkPlusOpenHapticsDevice::WriteConfiguration(vtkXMLDataElement* rootConfigElement)
+{
+  XML_FIND_DEVICE_ELEMENT_REQUIRED_FOR_WRITING(trackerConfig, rootConfigElement);
+  trackerConfig->SetAttribute("DeviceName", this->DeviceName.c_str());
+  return PLUS_SUCCESS;
+}
+
+PlusStatus vtkPlusOpenHapticsDevice::InternalStartRecording()
+{
+  if (!this->isHapticDeviceInitialized)
```
adamrankin: If these functions do nothing, omit them as the parent level does basic checks for you.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+HDCallbackCode HDCALLBACK
+vtkPlusOpenHapticsDevice::positionCallback(void* pData)
+{
+  
+  vtkPlusOpenHapticsDevice* client = reinterpret_cast<vtkPlusOpenHapticsDevice*>(pData);
+  HHD handle = client->DeviceHandle;
+  HD_state state = client->DeviceState;
+
+  const double unfilteredTimestamp = vtkPlusAccurateTimer::GetSystemTime();
+  ++client->FrameNumber;
```
adamrankin: Not really a big deal, but this will make it so that frame 0 never exists. Again, this is not wrong, just fyi.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  rasCorrection->SetElement(0, 0, -1);
+  rasCorrection->SetElement(1, 1, 0);
+  rasCorrection->SetElement(1, 2, 1);
+  rasCorrection->SetElement(2, 2, 0);
+  rasCorrection->SetElement(2, 1, 1);
+
+  //assemble force data
+  auto inputChannels = client->InputChannels;
+  vtkPlusDataSource* forceInput;
+  if (!inputChannels.empty() && inputChannels[0]->GetOwnerDevice()->GetToolByPortName("Force", forceInput))
```
adamrankin: Omit GetOwnerDevice, as you request a tool from a channel, not a device.

You are also guaranteed that InputChannels won't be empty, because you will have done that check in NotifiyConfigured
You can use client->InputChannels[0].GetTool(forceInput, "Force")

Also, be aware that most functions return PLUS_SUCCESS and PLUS_FAIL, so we generally explicitly compare to those values.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
sjh26: Thanks, will change this (and the related issues above).  This happened because I used the MicronTracker code as a starting template, and it is apparently out of date?

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  {
+    StreamBufferItem* item = new StreamBufferItem();
```
adamrankin: Replace with a stack variable and forceInput->GetLatestStreamBufferItem(&item)

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+    if (forceInput->GetLatestItemUidInBuffer() > 0)
+    {
+      forceInput->GetLatestStreamBufferItem(item);
```
adamrankin: Check that the return value is ITEM_OK, and also that the transform is valid with item->GetStatus()

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
sjh26: hdGetError() pops the most recent error off of its stack.  This is included because I have found that connection errors can persist in the stack for a surprisingly long time.  Old errors from a previous instance can prevent a new connection if the stack is not flushed.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  hdSetDoublev(HD_CURRENT_FORCE, force);
+  hdGetIntegerv(HD_CURRENT_BUTTONS, &buttonVals);
+  hdGetBooleanv(HD_CURRENT_INKWELL_SWITCH, &inkwell);
+  hdEndFrame(handle);
+
+  vtkPlusDataSource* stylus = NULL;
+  vtkPlusDataSource* velocity = NULL;
+  vtkPlusDataSource* buttons = NULL;
+
+  vtkSmartPointer<vtkTransform> transform = vtkSmartPointer<vtkTransform>::New();
```
adamrankin: If these are allocated and deallocated every update, might be worth it to make them member variables to optimize.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,297 @@
...
+  buttonMatrix->SetElement(1, 1, 0);
+  buttonMatrix->SetElement(2, 2, 0);
+  buttonMatrix->SetElement(0, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_1));
+  buttonMatrix->SetElement(1, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_2));
+  buttonMatrix->SetElement(2, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_3));
+  buttonMatrix->SetElement(3, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_4));
+  buttonMatrix->SetElement(0, 1, (int)inkwell);
+
+
+  if (client->GetToolByPortName("Stylus", stylus) == PLUS_SUCCESS)
```
adamrankin: You can simplify this a bit by just calling 
```
if(ToolTimeStampedUpdate("Stylus", ...) != PLUS_SUCCESS)
{
  LOG_ERROR(...)
}
```

src/PlusDataCollection/Testing/CMakeLists.txt
```diff
@@ -474,6 +474,13 @@ IF(PLUS_USE_OpenIGTLink)
     )
 ENDIF()
 
+#*************************** OpenHapticsDeviceTest *******************************
```
adamrankin: Yay tests!!! Thank you!

adamrankin:
Please also include a documentation page for the device (look for other .dox files in PlusLib, and the master device list page in PlusApp)

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
adamrankin: Ok! Sounds good to me.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
adamrankin: Yes, the code has been evolving for a long time. Not everything gets cleaned up to the latest "normal".

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
sjh26: I wanted to make it so the device could still run without a force input channel, so that's why I wanted to check the channels here.  Without the input channel I would just send a zero force vector to the device.

adamrankin: Oh, I see. Ok, where would the force data come from?

sjh26: The use case for this is integrating the haptic device with 3D slicer.  The force (as the translation in a transform) comes from Slicer broadcasting as an OpenIGTLinkTracker device (I have config files for both the force and no force cases).

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
sjh26: So, this accesses the tools by ID instead of portname, which does not seem to be working for me (can't find any of the tools based on their names in the config).  Same with client->InputChannels[0].GetTool(forceInput, "Force") you suggested above.  I have the IDs set in the configuration the same as the port name, but that internal ID seems to be something different?

sjh26: The ID that ToolTimeStampedUpdate is expecting is the transform name id (ex. StylusToTracker, etc).

adamrankin: Right, yes. Nevermind.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
sjh26: Is there a method to get a tool based on portname without going through the device object?  The channels seem to only support getting a tool based on the transform ID.

adamrankin: Accessor added! You can now get a tool by port name from a channel.

sjh26: awesome!

sjh26:
Made suggested fixes and added documentation

src/Documentation/UserManual/DeviceOpenHaptics.dox
```diff
@@ -0,0 +1,45 @@
...
+
+- \ref PackageWin64
+- \ref PackageLinux
+
+\section DeviceOpenHapticsInstallation Installation
+- Install OpenHaptics Developer SDK from: http://developer.geomagic.com/
+
+\section DeviceOpenHapticsConfigSettings Device configuration settings
+
+The device will output three transforms: the tracking transform of the the stylus,, the stylus velocity, and the stylus buttons.  In order to
```
adamrankin: Minor typo, double comma

src/Documentation/UserManual/DeviceOpenHaptics.dox
```diff
@@ -0,0 +1,45 @@
...
+transmit forces to the device, an input channel with a single tool DataSource (with portname "Force") should be provided.  The force vector should be stored in the
+translation portion of the transformation  matrix (RAS coordinates).
```
adamrankin: Minor typo, double space

src/PlusDataCollection/vtkPlusDeviceFactory.cxx
```diff
@@ -312,6 +316,10 @@ vtkPlusDeviceFactory::vtkPlusDeviceFactory()
   RegisterDevice("OpenCVVideo", "vtkPlusOpenCVCaptureVideoSource", (PointerToDevice)&vtkPlusOpenCVCaptureVideoSource::New);
 #endif
 
+#ifdef PLUS_USE_OPENHAPTICS
+  RegisterDevice("OpenHapticsDevice", "vtkPlusOpenHapticsDevice", (PointerToDevice)&vtkPlusOpenHapticsDevice::New);
```
adamrankin: Would you remove the Device suffix from the identifier (ie: just OpenHaptics)? Also matching identifier in .dox file

src/Documentation/UserManual/DeviceOpenHaptics.dox
```diff
@@ -0,0 +1,45 @@
...
+
+- \xmlElem \ref DataSources Three \c DataSource child elements are required. \RequiredAtt
+   - \xmlElem \ref DataSource \RequiredAtt
+   - \xmlAtt \ref PortName \RequiredAtt
+     - \c Stylus Tracking transform for stylus (RAS).
+     - \c StylusVelocity Stylus velocity measurement (RAS). The values are stored in the translation part of the transformation matrix. The rotation part is identity.
+     - \c Buttons States for buttons. The button values are stored in the first column of the matrix.  The inkwell switch is the first element in the second column.
+   - \xmlAtt \ref BufferSize \OptionalAtt{150}
+   - \xmlAtt \ref AveragedItemsForFiltering \OptionalAtt{20}
+- \xmlElem \ref InputChannelsList An Input channel is required to send force data to the device \RequiredAtt
```
adamrankin: I believe this tag is called InputChannels

src/Documentation/UserManual/DeviceOpenHaptics.dox
```diff
@@ -0,0 +1,45 @@
...
+   - \xmlElem \ref InputChannel \RequiredAtt
+     -\xmlElem \ref Id Identifier of an output channel of another device containing a tool with PortName="Force" \RequiredAtt
+
+
+\section DeviceOpenHapticsExampleConfigFile Example configuration file PlusDeviceSet_Server_OpenHaptics_WithForces.xml
+
+\include "ConfigFiles/PlusDeviceSet_Server_OpenHaptics_WithForces.xml"
```
adamrankin: Would you also include a link to your without forces example?

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,294 @@
...
+PlusStatus vtkPlusOpenHapticsDevice::InternalUpdate()
+{
+  //schedule the internal callback - must be done each time device is polled
+  //the scheduler is intended to manage multiple clients requesting data from the same device
+  hdScheduleSynchronous(positionCallback, this, HD_MAX_SCHEDULER_PRIORITY);
+
+  return PLUS_SUCCESS;
+}
+
+PlusStatus vtkPlusOpenHapticsDevice::InternalConnect()
```
adamrankin: Minor code formatting issue: would you add the 
`//----------------------------------------------------------------------------`
separators before each function?

ie:
```
...
}

PlusStatus vtkPlusOpenHapticsDevice::InternalConnect()
```

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,294 @@
...
+{
+  hdDisableDevice(DeviceHandle);
+  hdStopScheduler();
+  return PLUS_SUCCESS;
+}
+
+PlusStatus vtkPlusOpenHapticsDevice::ReadConfiguration(vtkXMLDataElement* rootConfigElement)
+{
+  XML_FIND_DEVICE_ELEMENT_REQUIRED_FOR_READING(deviceConfig, rootConfigElement);
+  XML_READ_CSTRING_ATTRIBUTE_REQUIRED(DeviceName,deviceConfig);
```
adamrankin: Minor code formatting, space between argument and comma

See Astyle formatter (if using Visual Studio, available as an extension) and Astyle formatting rules [here](https://github.com/PlusToolkit/PlusBuild/blob/master/Docs/BuildInstructionsWindows.md).

adamrankin: Also, since this is a std::string, you can use the XML_READ_STRING instead of CSTRING

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.cxx
```diff
@@ -0,0 +1,294 @@
...
+  client->buttonMatrix->SetElement(0, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_1));
+  client->buttonMatrix->SetElement(1, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_2));
+  client->buttonMatrix->SetElement(2, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_3));
+  client->buttonMatrix->SetElement(3, 0, (bool)(buttonVals & HD_DEVICE_BUTTON_4));
+  client->buttonMatrix->SetElement(0, 1, (int)inkwell);
+
+  if (client->GetToolByPortName("Stylus", stylus) == PLUS_SUCCESS)
+  {
+    client->ToolTimeStampedUpdate(stylus->GetId(), client->toolMatrix, TOOL_OK, client->FrameNumber, unfilteredTimestamp);
+
```
adamrankin: Minor code formatting, extra empty line

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+=========================================================Plus=header=end*/
+
+#ifndef __vtkPlusOpenHapticsDevice_h
+#define __vtkPlusOpenHapticsDevice_h
+
+#include "vtkPlusDataCollectionExport.h"
+
+#include "vtkPlusDevice.h"
+#include <HD/hd.h>
+#include "vtkTransform.h"
```
adamrankin: vtkTransform can be a forward declare and include in the .cxx file

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
+/*=Plus=header=begin======================================================
+  Program: Plus
+  Copyright (c) Laboratory for Percutaneous Surgery. All rights reserved.
+  See License.txt for details.
+=========================================================Plus=header=end*/
```
adamrankin: Please feel free to add your authorship information into the copyright area!

This is great work and much appreciated!

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include "vtkPlusDataCollectionExport.h"
+
+#include "vtkPlusDevice.h"
+#include <HD/hd.h>
+#include "vtkTransform.h"
+
+class vtkMatrix4x4;
+
+struct HD_state
```
adamrankin: If this is only used by this device, consider moving this into the device class to reduce global scope types.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+  /*! Disconnect from the device hardware */
+  PlusStatus InternalDisconnect();
+
+  /*! Read configuration from xml data */
+  virtual PlusStatus ReadConfiguration(vtkXMLDataElement* config);
+
+  /*! Read configuration from xml data */
+  virtual PlusStatus WriteConfiguration(vtkXMLDataElement* config);
+
+  vtkSetMacro(DeviceName, std::string);
```
adamrankin: We have created a vtkSetStdStringMacro(DeviceName) for this.

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+  vtkGetMacro(DeviceName, std::string);
```
adamrankin: Same with vtkGetStdStringMacro

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+protected:
+  vtkPlusOpenHapticsDevice();
+  ~vtkPlusOpenHapticsDevice();
+
+  
+  PlusStatus InternalStartRecording();
+  PlusStatus InternalStopRecording();
+
+  /*! Index of the last frame number. This is used for providing a frame number when the tracker doesn't return any transform */
+  double LastFrameNumber;
```
adamrankin: This seems like it should be an unsigned int as well?

src/PlusDataCollection/OpenHaptics/vtkPlusOpenHapticsDevice.h
```diff
@@ -0,0 +1,104 @@
...
+
+  unsigned int FrameNumber;
+
+  std::string DeviceName;
+  
+  bool isHapticDeviceInitialized;
```
adamrankin: Is this still a duplicate of some of the existing parent class variables?

adamrankin:
A few more minor fixes, but overall it's looking great.

I believe the Queen's group is preparing to make a release, so we will wait to merge this in until after the release is made (it will quickly become available in the nightly releases once committed).

One last thing is to ensure that any required DLLs are properly package for installation when this device is included in a package release, but I can fix that up after the commit is made (unless you'd like to take care of it, which is much appreciated). It is CMake code in the PlusDataCollection project.

sjh26:
Fixed issues, took  stab at the DLL thing.  

I also have branches for PlusLibData (config files) and PlusBuild (cmake changes, module for finding the OpenHaptics directories, etc) that go with this.  Let me know when i should make PRs for them.

adamrankin:
Building and testing pull requests one at a time, after the previous one is finished, I will incorporate this PR.

adamrankin:
Feel free to do the PlusLibData PR as that won't affect other people, and feel free to start the PlusBuild PR so we can review that as well.

Issue #249, opened 2017-10-31
PlusServer crashes when querying image list from StealthLink


lassoan:
When OpenIGTLinkRemote module is used in Slicer to query list of available images from StealthLink through Plus, PlusServer crashes.

Sunderlandkyl:
PlusServer shouldn't crash anymore when trying to query the image list or retrieve an image from StealthLink, however it still won't be functional until the openigtlink/OpenIGTLink#152 pull request is integrated.

Sunderlandkyl:
I may have spoken too soon. I was testing the StealthLink with a different dataset when it crashed while attempting to download the selected image.

Debugging now.

Sunderlandkyl:
I discovered that it was some sort of memory allocation issue in [vtkPlusStealthLinkTracker.cxx(669-670)](https://github.com/PlusToolkit/PlusLib/blob/master/src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx#L669-L670).
The server was crashing when trying to allocate ~223MB for the image data.

It was reproducible on a particular dataset for a while, unfortunately I am also not able to replicate the crash anymore.

It seems as though the issue resolved itself. Maybe it was caused by memory issues on my computer.

lassoan:
In a 32-bit application the usable memory space is quite limited. Allocating large contiguous memory block can be particularly problematic. Memory usage of PlusServer can be reduced by limiting size of image buffers (if there is any acquisition device). Also, we could enable ReleaseDataFlag (https://www.vtk.org/Wiki/VTK/FAQ#Using_ReleaseDataFlag) on the VTK filters that process the received image and/or instead of creating a pipeline, we would run filters manually (call Update(), GetOutput(), delete the producing filter).

Sunderlandkyl:
In order to reproduce the issue, I moved from the ~223MB CT to a ~523MB CT.
I haven't managed to fix it yet.

That has me wondering, at what point should we consider a dataset to be too large to support?

lassoan:
Theoretical maximum memory space available for a 32-bit process is 2GB. Practically it usually means that you have about 1GB for data. Allocating contiguous memory blocks may easily fail, even if you have enough memory, just not in one piece. I would say that if we can deal with CTs up to 512^3, 16 bits then it is enough. We should not crash, though, but report an error and keep running.

Sunderlandkyl:
This issue should now be fixed.

PR #266, opened 2017-11-08
re #249: Allow StealthLink to send larger images

Sunderlandkyl:
- ReleaseDataFlag is turned on for all filters when getting the image.
- Catch the bad_alloc exception that is thrown by VTK if an image is still to large, then display an error message and return failure.

The server will still crash if the image is too large in debug mode, however that may be unavoidable, since abort() is called in vtkGenericDataArray in debug mode.

Works on a 512^3 image using 2 bytes for each voxel in both release and debug.

Relevant issue: https://github.com/PlusToolkit/PlusLib/issues/249

### DIFF
--- a/src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
+++ b/src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
@@ -656,18 +656,29 @@ PlusStatus vtkPlusStealthLinkTracker::GetImage(const std::string& requestedImage
     examImageDirectoryToDelete = examImageDirectory;
 
     vtkSmartPointer<vtkDICOMImageReader> reader = vtkSmartPointer<vtkDICOMImageReader>::New();
-    reader->SetDirectoryName(examImageDirectory.c_str());
+    try
+    {
+      reader->SetDirectoryName(examImageDirectory.c_str());
+      reader->ReleaseDataFlagOn();
 
-    //to go from the vtk orientation to lps orientation, the vtk image has to be flipped around y and z axis
-    vtkSmartPointer<vtkImageFlip> flipYFilter = vtkSmartPointer<vtkImageFlip>::New();
-    flipYFilter->SetFilteredAxis(1);   // flip y axis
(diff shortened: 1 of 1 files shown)


lassoan:
Nice work!

src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -656,18 +656,29 @@ PlusStatus vtkPlusStealthLinkTracker::GetImage(const std::string& requestedImage
...
+      vtkSmartPointer<vtkImageFlip> flipZFilter = vtkSmartPointer<vtkImageFlip>::New();
+      flipZFilter->SetFilteredAxis(2);   // flip z axis
+      flipZFilter->ReleaseDataFlagOn();
+      flipZFilter->SetInputConnection(flipYFilter->GetOutputPort());
+      flipZFilter->Update();
+      imageData->DeepCopy(flipZFilter->GetOutput());
+    }
+    catch (const std::bad_alloc& e)
+    {
+      LOG_ERROR("Error when downloading image: " << e.what() << ". Image may be too large");
```
lassoan: Maybe add "from StealthLink"

    ...downloading the image from StealthLink: ...

src/PlusDataCollection/StealthLink/vtkPlusStealthLinkTracker.cxx
```diff
@@ -656,18 +656,29 @@ PlusStatus vtkPlusStealthLinkTracker::GetImage(const std::string& requestedImage
     examImageDirectoryToDelete = examImageDirectory;
 
     vtkSmartPointer<vtkDICOMImageReader> reader = vtkSmartPointer<vtkDICOMImageReader>::New();
```
lassoan: Is there a particular reason why the reader instantiation is outside the try/catch block?

Sunderlandkyl: The reader is still required later in the function, so I moved the declaration/instantiation outside the try/catch block.

Would it be better to have the declaration outside the try/catch, and move the instantiation back inside the block?

PR #267, opened 2017-11-09
Do not swamp the log with the exact same message

### LINKED ISSUE #256: Limit log file size
Is there a way to limit the size of the log file?

E.g. if a PlusServer loses connection to a device, a stream of messages is written to the log quickly (e.g. 30 times per second), making the size of the log file reach gigabyte proportions over a 24 hour period. Excerpt:
```
092817_143808.818|ERROR|067.533000| Unable to read position of second encoder with address: 3| in C:\P\Dzenan\PR17_x32\PlusLib\src\PlusDataCollection\USDigitalEncodersTracking\vtkPlusUSDigitalEncodersTracker.cxx(311)
092817_143808.853|ERROR|067.568000| Unable to read position of second encoder with address: 3| in C:\P\Dzenan\PR17_x32\PlusLib\src\PlusDataCollection\USDigitalEncodersTracking\vtkPlusUSDigitalEncodersTracker.cxx(311)
```

dzenanz:
A new macro is introduced which eases avoidance of repeating error messages in the log. It is applied in US Digital encoders, Capistrano ultrasound, and a few key places in PlusChannel and OpenIGTLinkServer. Closes #256.

### DIFF
--- a/src/PlusCommon/PlusCommon.cxx
+++ b/src/PlusCommon/PlusCommon.cxx
@@ -20,14 +20,14 @@ See License.txt for details.
 #include <string>
 
 //-------------------------------------------------------
-bool vtkPlusLogHelper::shouldWeLog(bool errorPresent)
+bool vtkPlusLogHelper::ShouldWeLog(bool errorPresent)
 {
   if (errorPresent)
   {
     ++m_Count;
     double timeStamp = vtkPlusAccurateTimer::GetSystemTime();
-    if (timeStamp - m_LastError > m_MinimumTimeBetweenLogging
-        || m_Count > m_MinimumCountBetweenLogging)
+    if (timeStamp - m_LastError > this->MinimumTimeBetweenLogging
+        || m_Count > this->MinimumCountBetweenLogging)
     {
       //log the error this time
       m_LastError = timeStamp;
@@ -41,7 +41,7 @@ bool vtkPlusLogHelper::shouldWeLog(bool errorPresent)
   }
   else //error has just been removed, reset to initial state
   {
-    m_LastError = -DBL_MAX / 2;
+    m_LastError = -std::numeric_limits<double>::max() / 2;
     m_Count = -2;
     return false;
   }
--- a/src/PlusCommon/PlusCommon.cxx
+++ b/src/PlusCommon/PlusCommon.cxx
@@ -26,8 +26,8 @@ bool vtkPlusLogHelper::ShouldWeLog(bool errorPresent)
   {
     ++m_Count;
     double timeStamp = vtkPlusAccurateTimer::GetSystemTime();
-    if (timeStamp - m_LastError > this->MinimumTimeBetweenLogging
-        || m_Count > this->MinimumCountBetweenLogging)
+    if (timeStamp - m_LastError > m_MinimumTimeBetweenLoggingSec
+        || m_Count > m_MinimumCountBetweenLogging)
     {
       //log the error this time
       m_LastError = timeStamp;
--- a/src/PlusCommon/PlusCommon.cxx
+++ b/src/PlusCommon/PlusCommon.cxx
@@ -20,6 +20,34 @@ See License.txt for details.
 #include <string>
 
 //-------------------------------------------------------
+bool vtkPlusLogHelper::shouldWeLog(bool errorPresent)
+{
+  if (errorPresent)
+  {
+    ++m_Count;
+    double timeStamp = vtkPlusAccurateTimer::GetSystemTime();
+    if (timeStamp - m_LastError > m_MinimumTimeBetweenLogging
+        || m_Count > m_MinimumCountBetweenLogging)
+    {
+      //log the error this time
+      m_LastError = timeStamp;
+      m_Count = 0;
+      return true;
+    }
+    else
+    {
+      return false;
+    }
+  }
+  else //error has just been removed, reset to initial state
+  {
+    m_LastError = -DBL_MAX / 2;
+    m_Count = -2;
+    return false;
+  }
+}
+
+//-------------------------------------------------------
 PlusTransformName::PlusTransformName()
 {
 }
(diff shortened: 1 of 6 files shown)


src/PlusCommon/PlusCommon.h
```diff
@@ -104,6 +104,30 @@ enum PlusImagingMode
   vtkPlusLogger::Instance()->LogMessage(logLevel, msgStream.str().c_str(), __FILE__, __LINE__); \
   }
 
+// If condition is satisfied, logs error only the first time
+// and returns PLUS_FAIL each time
+#define ERROR_FAIL_IF(condition, msg) \
```
adamrankin: This is great, but could we add a timeout feature?

In a long run, this error may occur on two separate unique instances, which would be nice to show two error messages.

Maybe 1 message per X minutes?

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -1141,11 +1141,8 @@ PlusStatus vtkPlusCapistranoVideoSource::InternalUpdate()
   GetObject(this->Internal->DataHandle, sizeof(BITMAP), &this->Internal->Bitmap);
 
   vtkPlusDataSource* aSource=NULL;
-  if( this->GetFirstActiveOutputVideoSource(aSource) != PLUS_SUCCESS )
-  {
-    LOG_ERROR("Unable to retrieve the video source in the ICCapturing device.");
-    return PLUS_FAIL;
-  }
+  ERROR_FAIL_IF(this->GetFirstActiveOutputVideoSource(aSource) != PLUS_SUCCESS,
```
adamrankin: Just an FYI, this should never fail as you can guarantee its existence in NotifyConfigured()

dzenanz: Why was it being checked then? I only re-wrote the check, I did not introduce it.

src/PlusCommon/PlusCommon.h
dzenanz: This can be done relatively easily. Will get to it!

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
adamrankin: The NotifyConfigured came in after a lot of these were written, and I did not go back and fix all instances as it's not a critical change. More it's just cleaning up a bit to only check once at post-config/pre-connect time for any configuration errors.

Again, optional, but good practice.

lassoan:
Calling `return` from the macro is very unexpected behavior. Either remove the `return` or rename the macro to something like RETURN_WITH_FAIL_IF(...).

Using a static bool to store if the message was printed already, is really ugly solution. In many cases it would not work, because you would need to restart the application to reset the flag.

Also have a look at the dynamic logging macro. It is added exactly for this purpose: to be able to dynamically suppress error messages temporarily (by logging at lower log level). See for how LOG_DYNAMIC is used for suppressing error messages in vtkPlusOpenIGTLinkServer.cxx.

Overall, if you want to make it very easy for developers to filter logs then you have to build larger and cleaner infrastructure behind it. For example, you could create a small class that:
- stores the last time when the condition occurred
- stores number of occurrences when the condition was true
- can tell if logging criteria is met (for example, x seconds elapsed or N number of occurrences)
- stores log levels for cases when logging criteria is met/not met (for example, by default it could be TRACE level when not met, ERROR level when met)
- can return log level that can be used in LOG_DYNAMIC.

src/PlusCommon/PlusCommon.cxx
```diff
@@ -20,6 +20,34 @@ See License.txt for details.
 #include <string>
 
 //-------------------------------------------------------
+bool vtkPlusLogHelper::shouldWeLog(bool errorPresent)
```
adamrankin: Minor formatting, functions are upper camel case

src/PlusCommon/PlusCommon.cxx
```diff
@@ -20,6 +20,34 @@ See License.txt for details.
...
+      return true;
+    }
+    else
+    {
+      return false;
+    }
+  }
+  else //error has just been removed, reset to initial state
+  {
+    m_LastError = -DBL_MAX / 2;
```
adamrankin: Could you use
std::numeric_limits<double>::max() instead?

src/PlusCommon/PlusCommon.h
```diff
@@ -59,6 +59,28 @@ enum PlusImagingMode
 ///////////////////////////////////////////////////////////////////
 // Logging
 
+class vtkPlusLogHelper
+{
+public:
+  double m_MinimumTimeBetweenLogging = 60.0;
```
adamrankin: Convention is no m_, and to use this-> in .cxx files

dzenanz: Is this only for public members? Otherwise e.g. `PlusTransformName::m_To` does not follow it.

lassoan: vtkPlus... classes are derived from vtkObject and use VTK conventions.
Non-VTK classes are names as Plus... and they use m_ prefix and no need for this->.

adamrankin: Huh, did not know that. Cool

lassoan: Include unit in variable name to make sure it it interpreted correctly: For example: m_MinimumTimeBetweenLogging => m_MinimumTimeBetweenLoggingSec

dzenanz:
More feedback or merge?

adamrankin:
Letting @lassoan manage the merges until the release is out.

dzenanz:
When is the release scheduled for?

Sunderlandkyl:
It should be done today.

dzenanz:
I guess this is the kind of change which could merged after :smile:

PR #278, opened 2017-11-23
Reducing log spam when 5dof tools are used, and 6th dof jumps all ove…

adamrankin:
…r the place

(example, optically tracked stylus with collinear fiducials)

### DIFF
--- a/src/PlusDataCollection/vtkPlusBuffer.cxx
+++ b/src/PlusDataCollection/vtkPlusBuffer.cxx
@@ -1331,7 +1331,11 @@ ItemStatus vtkPlusBuffer::GetInterpolatedStreamBufferItemFromTime(double time, S
   double angleDiffB = PlusMath::GetOrientationDifference(interpolatedMatrix, itemBmatrix);
   if (fabs(angleDiffA) > ANGLE_INTERPOLATION_WARNING_THRESHOLD_DEG && fabs(angleDiffB) > ANGLE_INTERPOLATION_WARNING_THRESHOLD_DEG)
   {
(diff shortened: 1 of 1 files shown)


lassoan:
The code change looks harmless, but you should not see jumps in the spinning angle when you move around a tool. Probably there is a single orientation of the stylus where these jumps may happen, and so you should set up your system in a way that you don't get into that specific orientation during normal use.

adamrankin:
Unfortunately with the 8700248 this situation happens constantly (collinear fiducials, NDI system is unable to reliably produce roll value). The net result is that the roll angle jumps randomly and produces this interpolation warning (as it should) constantly, flooding the log.

The solution is to use a different stylus or pattern, which is my final solution, but this helps me in development (and others is they want to use this aluminum machined stylus from NDI).

lassoan:
This should be ideally fixed in NDI software. Spinning can be restricted by always spin so that the spheres face "up" or any other chosen direction in some coordinate system. NDI Aurora 5-DOF sensors are stabilized like this. I don't know why they don't do it for this tracker/marker.

We could do this stabilization in Plus, too, but it would be simpler if they did it in their software because they know the sphere configuration, so they know what rotation axis should be constrained.

adamrankin:
That would be great, but won't likely happen. Imo it's not worth implementing in Plus, as there are more valuable things to accomplish. For now reducing the log spam is enough for me, and shouldn't affect any other cases where it happens for a genuine reason.

lassoan:
I agree. We cannot wait for NDI, don't have the time to implement instead of them, so a workaround is good enough (or don't use such tools or trackers).

PR #291, opened 2017-12-14
USDigitalEncoders xy position to use Plus Image coordinate system

jamesobutler:
I've updated the XY positioning math within USDigitalEncoders so that it uses the same coordinate system as the normal Plus image coordinate system.  Previously when capturing encoder position in the XZ plane and the image frame captured in XY, the Z coordinate had opposite signs between the two. This flips the Z coordinate to match.


dzenanz:
I don't think we need to change that in the code. What might be needed is change in the configuration, instead of `LocalAxis2="0 0 1"` have `LocalAxis2="0 0 -1"` (note the sign inversion).

jamesobutler:
Ok, I can do that instead.

PR #293, opened 2017-12-15
Image size typedef

adamrankin:
Well, this one hits a lot of files.

I did my best to find all uses of framesize, imagesize, etc...

For devices that don't require external SDKs, I think I've got everything. For devices that require SDKs, I did a search and replace approach, so there may be compiler errors/warnings.

### DIFF
--- a/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
+++ b/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
@@ -1461,9 +1461,9 @@ PlusStatus vtkPlusCapistranoVideoSource::SetBModeViewOption(unsigned int bModeVi
 }
 
 // ----------------------------------------------------------------------------
-PlusStatus vtkPlusCapistranoVideoSource::SetImageSize(int imageSize[3])
+PlusStatus vtkPlusCapistranoVideoSource::SetImageSize(const std::array<int, 3>& imageSize)
 {
-  return this->Internal->ImagingParameters->SetImageSize(imageSize, 3);
+  return this->Internal->ImagingParameters->SetImageSize(imageSize);
 }
 
 // ----------------------------------------------------------------------------
--- a/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
+++ b/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
@@ -34,11 +34,12 @@ public:
   std::vector<unsigned char>  MemoryBitmapBuffer;
   BITMAP                      Bitmap;
   bmBITMAPINFO                BitmapInfo;
-  BYTE                        *RfDataBuffer;
+  BYTE*                        RfDataBuffer;
   static const int            samplesPerLine = 2048;
 
   /* ! A data structure to include the parameters of Probe's Servo */
-  typedef struct {
+  typedef struct
+  {
     int                       JitterComp;
     int                       PositionScale;
     float                     SweepAngle;
@@ -48,7 +49,8 @@ public:
   } ProbeServo;
 
   /* ! A Combined data structure for ProbeParams */
-  typedef struct{
+  typedef struct
+  {
     ProbeType                 probetype;
     ProbeServo                probeservo;
     int                       Samples;
@@ -74,11 +76,11 @@ public:
   // ---------------------------------------------------------------------------
   // Public member functions  --------------------------------------------------
   /*! Constructor */
-  vtkPlusCapistranoVideoSource::vtkInternal::vtkInternal(vtkPlusCapistranoVideoSource * external)
-  : External( external )
-  , RfDataBuffer(NULL)
-  , ProbeHandle(NULL)
-  , ImagingParameters(vtkPlusUsImagingParameters::New())
+  vtkPlusCapistranoVideoSource::vtkInternal::vtkInternal(vtkPlusCapistranoVideoSource* external)
+    : External(external)
+    , RfDataBuffer(NULL)
+    , ProbeHandle(NULL)
+    , ImagingParameters(vtkPlusUsImagingParameters::New())
   {
     this->CreateUSProbeParamsDB();
     this->CreateUSProbePulserParamsDB();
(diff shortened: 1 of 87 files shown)


dzenanz:
With a move to C++11, this is a logical and welcome change!

adamrankin:
A review would be most welcome. I'm sure I've missed some parts, and there were some changes in the Capistrano class.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -683,15 +685,15 @@ PlusStatus vtkPlusCapistranoVideoSource::WriteConfiguration(vtkXMLDataElement* r
...
     LOG_WARNING("vtkPlusIntersonVideoSource is expecting one output channel and there are "
-                 << this->OutputChannels.size() << " channels. First output channel will be used.");
+                << this->OutputChannels.size() << " channels. First output channel will be used.");
   }
 
-  if( this->OutputChannels.empty() )
+  if (this->OutputChannels.empty())
   {
-    LOG_ERROR("No output channels defined for vtkPlusIntersonVideoSource. Cannot proceed." );
+    LOG_ERROR("No output channels defined for vtkPlusIntersonVideoSource. Cannot proceed.");
```
dzenanz: This should be vtkPlusCapistranoVideoSource. I guess Capistrano wrapper was based on Interson one :smile:

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -866,24 +868,23 @@ PlusStatus vtkPlusCapistranoVideoSource::SetupProbe(int probeID)
...
   if (display == NULL)
   {
     LOG_ERROR("Could not initialize the display");
     return PLUS_FAIL;
   }
 
   // Initialize DIB -----------------------------------------------------------
-  int frameSize[2] = {imageSize[0], imageSize[1]};
-  this->Internal->InitializeDIB(frameSize);
+  this->Internal->InitializeDIB(imageSize);
```
dzenanz: This seems to be the most significant change.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -1589,20 +1589,20 @@ PlusStatus vtkPlusCapistranoVideoSource::SetDepthMm(float depthMm)
...
 }
 
 // ----------------------------------------------------------------------------
 PlusStatus vtkPlusCapistranoVideoSource::UpdateUSProbeParameters()
 {
   // Set US Probe scan mode -------------------------------------------------
-  if(this->BidirectionalMode)
-    usbSetBidirectionalMode();
+  if (this->BidirectionalMode)
+  { usbSetBidirectionalMode(); }
```
dzenanz: Doesn't the style call for braces on lines by themselves to emphasize an if? This is hardly better, if not even worse than no braces.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.h
```diff
@@ -270,8 +269,8 @@ class vtkPlusDataCollectionExport vtkPlusCapistranoVideoSource: public vtkPlusUs
...
 
 private:
-  vtkPlusCapistranoVideoSource(const vtkPlusCapistranoVideoSource &); // Not implemented
-  void operator=(const vtkPlusCapistranoVideoSource &); // Not implemented
+  vtkPlusCapistranoVideoSource(const vtkPlusCapistranoVideoSource&);  // Not implemented
+  void operator=(const vtkPlusCapistranoVideoSource&);  // Not implemented
 };
 
-#endif
+#endif
```
dzenanz: Perhaps add a newline at the end?

dzenanz:
@jperdomo23 @jamesobutler Can you test this PR?

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
adamrankin: Yes, my astyle formatter was not set up correctly. It should place them in brackets on their own lines. I will correct.

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.h
dzenanz: https://stackoverflow.com/a/37684763/276168

adamrankin: Aaaahhh... I will stop removing them!

jamesobutler:
Sure, I can test the capistrano class once changes are complete.

adamrankin:
@lassoan @cpinter I'd like to merge this (I am testing my current master merge now), before more commits make it harder to merge.

I've done my best, but suspect I'll still have broken some things. I have time to address them, so is now an ok time to merge (after merging in master to this branch)?

lassoan:
As I wrote above, I'm OK with merging this.

adamrankin:
Ok, running tests now and then will merge.

cpinter:
Looks good to me, thanks! What is the main reason for switching to std::array?

adamrankin:
Personal preference. I wanted a std type so that all of the std algorithms
could work on it if necessary.

I also wanted a fixed size of 3, to reduce the amount of if(size == 2)
etc... Only the devices need set the third element to 1 and never touch it
again (for 2d devices).

cpinter:
@adamrankin Yes the fixed size of 3 makes complete sense to me. I was curious about all the other variables you changed. Thanks!

adamrankin:
I guess I tend to prefer types like std::array<int, 3> over int* or int[3].

They seem less prone to ambiguity or misintepretation. I know good variable naming will help identify the meaning of a *, but std::array<int, 3> is always clear (in my opinion).

adamrankin:
Plus, std::array<int, 3> as a return type will be returned by copy, meaning less error checking on return values.

Issue #283, opened 2017-12-06
CMake 3.10.0 automoc generates files in a different directory


markasselin:
From the CMake 3.10.0 release notes:
"When using AUTOMOC or AUTOUIC with a multi configuration generator (e.g. Xcode), included *.moc, moc_*.cpp and ui_*.h files are generated in <AUTOGEN_BUILD_DIR>/include_<CONFIG> instead of <AUTOGEN_BUILD_DIR>/include."

@lassoan this was an intentional change.

lassoan:
Could we add both .../include and  .../include_ to the list of include directories?

markasselin:
Unfortunately not - this leaves non-existent directories in the INTERFACE_INCLUDE_DIRECTORIES. I have a working solution which wraps the addition of .../include, .../include_Debug and .../include_Release in if statements checking if the directory exists - but there must be a cleaner solution than this...

lassoan:
It is possible that these include directories are set as target properties in some targets. If we link that target then the appropriate include directory is added automatically. Could you check if build succeeds if we simply remove the manual directory include?

markasselin:
No luck, build fails without the manual directory include.

lassoan:
Nice work! Don't forget to close this issue if the problem is resolved.

markasselin:
@lassoan Just a reminder to merge [PR 284](https://github.com/PlusToolkit/PlusLib/pull/284)

lassoan:
Thank you!

PR #294, opened 2017-12-16
re #283: Fixing CMake update to build first time with 3.8 and above

markasselin:
Tested and confirmed successful first time builds for CMake 3.7.2, 3.9.6, 3.10.0 and newly release 3.10.1.

### DIFF
--- a/src/PlusWidgets/CMakeLists.txt
+++ b/src/PlusWidgets/CMakeLists.txt
@@ -45,16 +45,14 @@ SET(${PROJECT_NAME}_INCLUDE_DIRS
 IF(NOT (CMAKE_VERSION VERSION_LESS "3.8"))
   # CMake 3.8 and later puts generated headers (such as ui_QPlusDeviceSetSelectorWidget.h) into
   # .../PlusLib-bin/src/PlusWidgets/PlusWidgets_autogen/include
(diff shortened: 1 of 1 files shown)


src/PlusWidgets/CMakeLists.txt
```diff
@@ -45,16 +45,14 @@ SET(${PROJECT_NAME}_INCLUDE_DIRS
 IF(NOT (CMAKE_VERSION VERSION_LESS "3.8"))
   # CMake 3.8 and later puts generated headers (such as ui_QPlusDeviceSetSelectorWidget.h) into
   # .../PlusLib-bin/src/PlusWidgets/PlusWidgets_autogen/include
-  IF(IS_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}_autogen/include)
-    LIST(APPEND ${PROJECT_NAME}_INCLUDE_DIRS ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}_autogen/include)
-  ENDIF()
-  
+  file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}_autogen/include)
```
adamrankin: Random question, is the make_directory necessary?

markasselin: I tried this without it - and if the directory doesn't exist CMake will fail with the error "include directories not found".

PR #297, opened 2017-12-18
Enable a configurable option to toggle send blocking on client sockets.

adamrankin:
This prevents disconnects due to insufficient bandwidth on a network. Hopefully no longer necessary if image compression sending is implemented.

### DIFF
--- a/src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
+++ b/src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
@@ -16,6 +16,7 @@ PlusIgtlClientInfo::PlusIgtlClientInfo()
   , Resolution(0)
   , TDATARequested(false)
   , LastTDATASentTimeStamp(-1)
+  , SendBlocking(false)
 {
 
 }
@@ -53,12 +54,19 @@ PlusStatus PlusIgtlClientInfo::SetClientInfoFromXmlData(vtkXMLDataElement* xmlda
 
   if (xmldata->GetAttribute("TDATARequested") != NULL)
   {
-    clientInfo.TDATARequested = STRCASECMP(xmldata->GetAttribute("TDATARequested"), "TRUE") == 0;
+    clientInfo.SetTDATARequested(STRCASECMP(xmldata->GetAttribute("TDATARequested"), "TRUE") == 0);
   }
 
   if (xmldata->GetAttribute("Resolution") != NULL)
   {
-    xmldata->GetScalarAttribute("Resolution", clientInfo.Resolution);
+    int resolution;
+    xmldata->GetScalarAttribute("Resolution", resolution);
+    clientInfo.SetResolution(resolution);
+  }
+
+  if (xmldata->GetAttribute("SendBlocking") != NULL)
+  {
+    clientInfo.SetSendBlocking(STRCASECMP(xmldata->GetAttribute("SendBlocking"), "TRUE") == 0);
   }
 
   // Get message types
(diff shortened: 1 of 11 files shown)


src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
```diff
@@ -53,12 +54,19 @@ PlusStatus PlusIgtlClientInfo::SetClientInfoFromXmlData(vtkXMLDataElement* xmlda
 
   if (xmldata->GetAttribute("TDATARequested") != NULL)
   {
-    clientInfo.TDATARequested = STRCASECMP(xmldata->GetAttribute("TDATARequested"), "TRUE") == 0;
+    clientInfo.SetTDATARequested(STRCASECMP(xmldata->GetAttribute("TDATARequested"), "TRUE") == 0);
   }
 
   if (xmldata->GetAttribute("Resolution") != NULL)
```
lassoan: Is there a reason not using the usual config XML macros?

lassoan: "Resolution" is a really bad parameter name and it is not even documented. Could you please rename it and add documentation (see description here: https://github.com/openigtlink/OpenIGTLink/blob/master/Documents/Protocol/trackingdata.md). For backward compatibility, allow using the old attribute name as well, but log a warning.

src/PlusOpenIGTLink/PlusIgtlClientInfo.h
```diff
@@ -63,21 +72,28 @@ struct vtkPlusOpenIGTLinkExport PlusIgtlClientInfo
...
   std::vector<ImageStream> ImageStreams;
 
-  /*! A new TDATA is only sent if the time elapsed is at least the resolution
-     value (otherwise we don't send this tracking data to the client) */
-  int Resolution;
+protected:
+  /*! IGTL header version supported by the client */
+  int ClientHeaderVersion;
+
+  /*! Flag for send blocking or not (default no) */
```
lassoan: This description is not clear. What does it do exactly?
Also, if there are public get/set functions then the value should be documented there.

src/PlusOpenIGTLink/igtlPlusClientInfoMessage.h
```diff
@@ -2,54 +2,52 @@
...
+  */
+  class vtkPlusOpenIGTLinkExport PlusClientInfoMessage: public StringMessage
+  {
+  public:
+    typedef PlusClientInfoMessage          Self;
+    typedef StringMessage                  Superclass;
+    typedef SmartPointer<Self>             Pointer;
+    typedef SmartPointer<const Self>       ConstPointer;
+
+    igtlTypeMacro(igtl::PlusClientInfoMessage, igtl::StringMessage);
```
lassoan: GitHub's web interface does not have the option of ignoring whitespace changes. Therefore, if you mix real changes with large amount of whitespace changes then it makes the review much more difficult. So, in the future please try to separate code style updates from functional changes.

src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
adamrankin: No, just updated existing code. Will convert to macros.

adamrankin:
Agreed, SetSendBlocking is a convenience function, but it's more descriptive than sendtimeout to 0, in my opinion.

src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
adamrankin: Agreed, fyi resolution is the name of the parameter for the TDATA message.

lassoan:
Adding this parameter as is, could cause complications much more difficult than accepting the fact that 0 means no timeout. The most obvious error case is that you set "DefaultClientSendTimeoutSec" attribute and it just does not work, because there is another attribute with a completely different name ("SendBlocking") that you don't know and don't care about that interferes with this. If the attributes have very similar names, for example "DefaultClientSendTimeoutSec" and "DefaultClientSendTimeoutEnabled" then user would a higher chance of figuring this out, but overall I feel that adding two attributes for setting a single parameter has potential for causing trouble while offering very little benefit.

adamrankin:
Interesting. Some of the default client info comes from the ClientInfo tag, and some comes from the parent server attributes.

Fair enough, I will remove the send blocking attribute, and document that 0 means no timeout (if not already done).

PR #307, opened 2018-01-01
Save custom fields to file

dzenanz:


src/PlusDataCollection/vtkPlusBuffer.cxx
```diff
@@ -958,6 +958,15 @@ PlusStatus vtkPlusBuffer::WriteToSequenceFile(const char* filename, bool useComp
     frameNumberFieldValue << std::fixed << frameNumber;
     trackedFrame->SetCustomFrameField("FrameNumber", frameNumberFieldValue.str());
 
+    // Add custom fields
+    const PlusTrackedFrame::FieldMapType& customFields = bufferItem.GetCustomFrameFieldMap();
+    PlusTrackedFrame::FieldMapType::const_iterator cf = customFields.begin();
+    while (cf != customFields.end())
```
lassoan: Is there a particular reason you use `while` instead of `for` loop as everywhere else?
When we switch to c++11 (we just wait for StealthLink now) format of for loop will be very clean and simple.

dzenanz: I don't know whether I could use `auto` there, and without it the declarations would be longer, and hence less readable. That is the only reason.

lassoan: OK. In general, if I had to choose between slightly improved syntax of being consistent with the rest of the code, I would prefer consistency to make the code easier to understand and maintain.

In this case the syntax is not really simpler and we'll not find the code when we'll revise all `for` statements. If it's easy for you to update the code to the usual for loop then please do it. If not, then let us know and we'll merge it as is (and maybe later update it). Thanks!

adamrankin:
Hey @dzenanz, I hope it's ok I went ahead and switched it to a for loop.

dzenanz:
I was a bit sick the last two days (some throwing up and temporary loss of appetite), so I didn't do it myself. Switching to a for loop is fine.

PR #317, opened 2018-01-16
Fix VTK deprecated code

samuelgerber:
PlusLib failed to compile on Ubuntu due to use of deprecated VTK macros / header files


adamrankin:
Thanks, taking a quick look!

PR #341, opened 2018-03-08
Store server command response messages in attributes, instead of between tags

Sunderlandkyl:
It was already handled this way for old (V1/V2) OpenIGTLink commands.
This changes the current current V3 commands to use the a similar message format.
Also adds command name and status attributes to response.

### DIFF
--- a/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
+++ b/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
@@ -1301,13 +1301,15 @@ igtl::MessageBase::Pointer vtkPlusOpenIGTLinkServer::CreateIgtlMessageFromComman
 
       // Send command result details both in XML and in metadata, slowly phase towards metadata
       std::ostringstream replyStr;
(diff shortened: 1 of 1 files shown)


adamrankin:
This changes the external api. Do you know where this is used/handled? SlicerIGT maybe?

adamrankin:
This also changes the content of the result variable from true/false to SUCCESS/FAIL

adamrankin:
Ah, I see, this is how SlicerIGT was expecting the format. Is this XML structure documented anywhere? If so, could you include the link in this PR

Sunderlandkyl:
SlicerIGT uses the v1/v2 commands in PlusRemote. I'm in the process of migrating it to use the v3 commands.
I don't know of anything else that uses the v3 commands.

I had originally left the SUCCESS/FAIL as true/false, but changed it to make format more similar to the v1/v2 messages. I think the true/false is probably still better.

Sunderlandkyl:
I don't recall seeing the XML structure documented anywhere.

adamrankin:
@adamrankin reminder to update uwpopenigtlink code

adamrankin:
Maybe it's time to have a v3 page on https://github.com/openigtlink/openigtlink for the command message.

Issue #172, opened 2017-09-21
Create PLUS device to enhance spinal ultrasound images and detect bone surfaces


Sunderlandkyl:
(No description)

Migrated from https://app.assembla.com/spaces/plus/tickets/1108/details

Sunderlandkyl:
2016-08-23 09:58
Ben Church
re #1108 Added vtkPlusTransverseProcessEnhancer device which currently just performs a vtkImageGaussianSmooth on the input ultrasound image. Added config file for this device and modified vtkP
Committed to: plus
In plus:4823

2016-08-25 12:48
Ben Church
re #1108: Added threshold operation to image processing pipeline. Image processor now reads more parameters from config file, making experimentation faster.
Committed to: plus
In plus:4831

2016-08-26 16:40
Adam Rankin
re #1108 Updating code to adhere to coding conventions. Fixing bugs relating to NULL checks. Implementing complete WriteConfiguration function.
Committed to: plus
Updating last of the PlusCommon::Trim usages
In plus:4834

2016-09-08 14:41
Ben Church
re #1108: Implemented Sobel edge detection plus a function to extract and combine its two-component output. I will next investigate pre-existing filters for component combination, island remov
Committed to: plus
In plus:4848

2016-09-09 12:29
Adam Rankin
re #1108 Reducing Mac build warnings
Committed to: plus
Updating code to adhere to coding conventions

Updating WriteConfiguration to output all stored values

Updating config file name to match convention
In plus:4850

2016-09-09 12:32
Adam Rankin
re #1108 Fixing linux build error
Committed to: plus
In plus:4851

2016-09-14 13:51
Ben Church
re #1108: Implemented another thresholder to binarize image so that vtkIslandRemoval2D works.
Committed to: plus
In plus:4867

2016-09-14 14:12
Ben Church
re #1108: Reverted since my last commit (or some attempt at conflict resolution) deleted NVidiaDVP files.
Committed to: plus
In plus:4868

2016-09-14 14:24
Ben Church
re #1108: Second commit attempt - Implemented another thresholder to binarize image so that vtkIslandRemoval2D works.
Committed to: plus
In plus:4869

2016-09-14 15:10
Ben Church
re #1108: Implemented "ImageConjunction" to retrieve pixel values from before image was binarized.
Committed to: plus
In plus:4870

2016-09-21 21:24
Ben Church
re #1108: Implemented image erosion and dilation as separate processes
Committed to: plus
In plus:4886

[...]

PR #343, opened 2018-03-14
ENH re #172: Implemented saving intermediate images 

BenChurch:
@lassoan @cpinter @ungi 
I think this commit with intermediate image saving for vtkPlusBoneEnhancer is ready to merge. Although I encountered a separate problem trying to implement this for vtkPlusTransverseProcessEnhancer:

I tried testing the vtkPlusTransverseProcessEnhancer device with the same image saving changes by setting the Processor config elements Type attribute to Type="vtkPlusTransverseProcessEnhancer" but running it as a Plus device returned the error:
|ERROR|271.682000|SERVER> Unknown processor type: vtkPlusTransverseProcessEnhancer| in :\D\plusGit\PB-B\PlusLib\src\PlusDataCollection\ImageProcessor\vtkPlusImageProcessorVideoSource.cxx(128)

Is anything obviously wrong with vtkPlusTransverseProcessEnhancer's type name to any of you?

I understand that vtkPlusTransverseProcessEnhancer inherits functionality from vtkPlusBoneEnhancer since they were seperated in [https://github.com/PlusToolkit/PlusLib/commit/fe12a12ccd86ad7d9336b1dce584c6dc88b53580](url) and [https://github.com/adamrankin/PlusLib/commit/c1965919c8bdc732c4ff74c7cb6d3cff42eb56ba](url). I do not understand the full extent of the inheritance if it relates to the type name.

I will begin investigating unless someone knows the solution right off hand.

### DIFF
--- a/src/PlusImageProcessing/vtkPlusBoneEnhancer.cxx
+++ b/src/PlusImageProcessing/vtkPlusBoneEnhancer.cxx
@@ -55,8 +55,11 @@ vtkPlusBoneEnhancer::vtkPlusBoneEnhancer()
 
   LinesImage(NULL),
   ProcessedLinesImage(NULL),
-  FirstFrame(true)
+  FirstFrame(true),
+
+  SaveIntermediateResults(false)
 {
+
   this->GaussianSmooth = vtkSmartPointer<vtkImageGaussianSmooth>::New();    // Used to smooth the image
   this->EdgeDetector = vtkSmartPointer<vtkImageSobel2D>::New();             // Used to outline edges of the image
   this->ImageBinarizer = vtkSmartPointer<vtkImageThreshold>::New();         // Used to convert into a binary image
@@ -168,6 +171,17 @@ PlusStatus vtkPlusBoneEnhancer::ReadConfiguration(vtkSmartPointer<vtkXMLDataElem
   if (imageProcessingOperations != NULL)
   {
 
+    // Read SaveIntermediateResults tag
+    vtkSmartPointer<vtkXMLDataElement> saveIntermediateResultsBool = imageProcessingOperations->FindNestedElementWithName("SaveIntermediateResults");
+    if (saveIntermediateResultsBool == NULL)
+    {
+      LOG_WARNING("Unable to locate SaveIntermediateResults element. Using default value: " << std::boolalpha << (this->SaveIntermediateResults));
+    }
+    else
+    {
+      XML_READ_BOOL_ATTRIBUTE_OPTIONAL(SaveIntermediateResults, saveIntermediateResultsBool);
+    }
+    
     // Read tags related to the Gaussian filter
     vtkSmartPointer<vtkXMLDataElement> gaussianParameters = imageProcessingOperations->FindNestedElementWithName("GaussianSmoothing");
     if (gaussianParameters == NULL)
@@ -250,6 +264,9 @@ PlusStatus vtkPlusBoneEnhancer::WriteConfiguration(vtkSmartPointer<vtkXMLDataEle
 
   //Write the parameters for filters to the output config file
   XML_FIND_NESTED_ELEMENT_CREATE_IF_MISSING(imageProcessingOperations, processingElement, "ImageProcessingOperations");
+
+  XML_FIND_NESTED_ELEMENT_CREATE_IF_MISSING(saveIntermediateResultsBool, imageProcessingOperations, "SaveIntermediateResults");
+  XML_WRITE_BOOL_ATTRIBUTE(SaveIntermediateResults, saveIntermediateResultsBool)
 
   XML_FIND_NESTED_ELEMENT_CREATE_IF_MISSING(gaussianParameters, imageProcessingOperations, "GaussianSmoothing");
   gaussianParameters->SetDoubleAttribute("GaussianStdDev", this->GaussianStdDev);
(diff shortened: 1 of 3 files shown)


src/PlusImageProcessing/vtkPlusBoneEnhancer.cxx
```diff
@@ -168,6 +170,17 @@ PlusStatus vtkPlusBoneEnhancer::ReadConfiguration(vtkSmartPointer<vtkXMLDataElem
   if (imageProcessingOperations != NULL)
   {
 
+    // Read SaveIntermediateResults tag
+    vtkSmartPointer<vtkXMLDataElement> SaveIntermediateResultsBool = imageProcessingOperations->FindNestedElementWithName("SaveIntermediateResults");
```
cpinter: Local variables with lower case please: saveIntermediateResultsBool
The way we differentiate between member and local variables is the case of the first letter, so this is good to keep consistent to increase readability.

src/PlusImageProcessing/vtkPlusBoneEnhancer.cxx
```diff
@@ -52,11 +52,13 @@ vtkPlusBoneEnhancer::vtkPlusBoneEnhancer()
   IslandAreaThreshold(-1),
   BoneOutlineDepthPx(3), // Note: this only changes the appearance/thickness of the 3D model. Different numbers do not change what is or is not marked as bone.
   BonePushBackPx(9),     // Horizontal distance between where a shadow is located, and where the bone begins
-
+
```
cpinter: Remove the trailing whitespaces please

cpinter:
I added a few comments. Also I noticed the following:
This PR includes 2 commits. In the first one you remove some commented code. And in the second one you add it back! Please do something about this. Either don't add it back, or don't remove it.

cpinter:
About the error, I think is due to a missing specified. Bryan should have added (and I sohuld have noticed that he didn't) a line like this
https://github.com/PlusToolkit/PlusLib/blob/master/src/PlusImageProcessing/vtkPlusBoneEnhancer.h#L49
in the child class

cpinter:
Of course besides my nitpicky points your changes are good to go :)

Please add the GetProcessorTypeName function in vtkPlusTransverseProcessEnhancer as well. Good catch though!

BenChurch:
I see, I'll write up these changes. Squashing the commits is probably a good idea, I thought the second commit without the comments removed overwrote the first. I've had trouble with squashing but will update the pull request when I squash all the changes.

cpinter:
You can simply open the log in TortoiseGit, select the two commits, and click on Squash commits or similar.

cpinter:
@BenChurch The latest pushes has further complicated things, and I think it wouldn't compile either because merge conflict indicators are in the code now. You'll need to clean it up. Please don't do more commits, it just makes things worse.

What I'd suggest doing is
1. Exporting your changes to a folder (TortoiseGit log -> select modified files -> right-click -> Export files - may need to do this for more than one commits, make sure you have all changed files!)
2. Hard reset your branch (again TortoisGit log, right-click the last commit that was before yours, right-click -> reset -> hard)
3. Copy the changes from the folder you exported the files to PlusLib
4. Make the commit again
5. Push -> ignore unknown changes

If you need help you can call me on skype, or if you want I can do this for you.

cpinter:
@BenChurch also iIn general (not in this case, we're beyond that point now) I'd like to bring your attention to the "amend commit" command, that allows modifying a commit without adding more and more. Usually this is what the developers do when we make changes according to comments on a PR. (If you push an amended commit, then you'll need to use the "ignore unknown changes" switch so that it goes through, because it overwrites the history)

src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
```diff
@@ -1313,7 +1313,6 @@ igtl::MessageBase::Pointer vtkPlusOpenIGTLinkServer::CreateIgtlMessageFromComman
       replyStr << " Message=\"" << commandResponse->GetResultString() << "\"></CommandReply>";
       igtlMessage->SetMetaDataElement("Message", IANA_TYPE_US_ASCII, commandResponse->GetResultString());
 
-      igtlMessage->SetMetaDataElement("Result", IANA_TYPE_US_ASCII, (commandResponse->GetStatus() ? "true" : "false"));
```
cpinter: Was this change intentional?

BenChurch: No, I don't know how that got deleted. I think I fixed it in the latest commit.

cpinter:
You know what, don't worry I'll clean this up, it will take like 15 minutes to do it right.

One last suggestion if I may: when you issue a PR, it's good to make it from a topic branch even if you did the work in the master branch of your working copy.

BenChurch:
I was able to follow your instructions so I think everything is in order.

cpinter:
Indeed, it looks good, great!

I'll add one more comment in the code in a second, and please add the GetProcessorTypeName function in vtkPlusTransverseProcessEnhancer as well. Please make sure you do amend commit (in TortoiseGit you simply check the Amend checkbox in the Commit window, then the ignore unknown changes checkbox in the Push window). That will be all. Thanks a lot!

src/PlusImageProcessing/vtkPlusBoneEnhancer.cxx
```diff
@@ -57,6 +57,8 @@ vtkPlusBoneEnhancer::vtkPlusBoneEnhancer()
   ProcessedLinesImage(NULL),
   FirstFrame(true)
 {
+  this->SetSaveIntermediateResults(false);
```
cpinter: Please initialize this variable a few lines above, like the other member variables.

cpinter:
I know the GitHub process looks complicated, but once you get it working, it'll be a breeze I promise you :)

BenChurch:
Yeah there's a lot of functionality to mess things up with. This Amend tool does seem handy though.

BenChurch:
Thanks, the Amend and ignore unknown changes are definitely the way to do this. All the changes should be there.

BenChurch:
I don't think I have the write access needed to merge this. Would you mind doing that?

cpinter:
Sure. Integrated in https://github.com/PlusToolkit/PlusLib/commit/ea13c2bff7983aee321a1e8273105f2752844011

PR #373, opened 2018-06-28
Add GetUsParameter command and change format of SetUsParameter response

Sunderlandkyl:
Sent GetUsParameter:
```
<Command Name="GetUsParameter" DeviceId="VideoDevice">
  <Parameter Name="DepthMm"/>
  <Parameter Name="FrequencyMhz"/>
  <Parameter Name="GainPercent"/>
  <Parameter Name="TEST"/>
</Command>
```
GetUsParameter response:
```
<CommandReply>
  <Parameter Name="DepthMm" Success="true" Value="35"/>
  <Parameter Name="FrequencyMhz" Success="true" Value="3.75"/>
  <Parameter Name="GainPercent" Success="true" Value="50"/>
  <Parameter Name="TEST" Success="false"/>
</CommandReply>
```
Response metadata:
````
DepthMm=35
Error=Invalid parameter TEST. 
FrequencyMhz=3.75
GainPercent=50
Status=FAIL
````

Sent SetUsParameter:
```
<Command Name="SetUsParameter" DeviceId="VideoDevice">
  <Parameter Name="DepthMm" Value="35"/>
  <Parameter Name="GainPercent" Value="50"/>
  <Parameter Name="ABCDEFG" Value="50"/>
</Command>
```
SetUsParameter response:
```
<CommandReply>
  <Parameter Name="ABCDEFG" Success="false"/>
  <Parameter Name="DepthMm" Success="true"/>
  <Parameter Name="GainPercent" Success="true"/>
</CommandReply>
```
Response metadata:
````
ABCDEFG=FAIL
DepthMm=SUCCESS
Error=Invalid parameter name ABCDEFG
GainPercent=SUCCESS
Status=FAIL
````
___
**OLD FORMAT:**
~I quickly threw together a GetUsParameter command based on the SetUsParameter command.~

~**Sent command:**~
```
<Command Name="GetUsParameter" DeviceId="VideoDevice">
  <Parameter Name="DepthMm"/>
  <Parameter Name="FrequencyMhz"/>
  <Parameter Name="GainPercent"/>
  <Parameter Name="TEST"/>
</Command>
```
~**Response:**~
```
<CommandReply
  Name="GetUsParameter" Status="FAIL"
  Error="Parameter: DepthMm - not set. Invalid parameter name: TEST.  " 
  Message="FrequencyMhz=3.56,GainPercent=52,">
</CommandReply>
```
~The available requested parameters are also attached to the command as metadata:~
```
Metadata:
    Error=Parameter: DepthMm - not set. Invalid parameter name: TEST. 
    FrequencyMhz=3.56
    GainPercent=52
    Message=FrequencyMhz=3.56,GainPercent=52,
    Status=FAIL
```

### DIFF
--- a/src/PlusDataCollection/vtkPlusUsImagingParameters.cxx
+++ b/src/PlusDataCollection/vtkPlusUsImagingParameters.cxx
@@ -418,8 +418,12 @@ PlusStatus vtkPlusUsImagingParameters::ReadConfiguration(vtkXMLDataElement* devi
   for (int i = 0; i < parameterList->GetNumberOfNestedElements(); ++i)
   {
     vtkXMLDataElement* element = parameterList->GetNestedElement(i);
-    std::string name(element->GetAttribute("name"));
-    std::string value(element->GetAttribute("value"));
+    std::string name = element->GetAttribute("name") ? element->GetAttribute("name") : "";
+    std::string value = element->GetAttribute("value") ? element->GetAttribute("value") : "";
+    if (name.empty())
+    {
+      continue;
+    }
 
     if (this->Parameters[name].Value != value)
     {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by Sunderlandkyl** [comment]
Neither the SetUsParameter or GetUsParameter commands follow the convention described here: http://igsio.github.io/.

Is that the standard that should be used?

**Comment by adamrankin** [comment]
Yes, if the standard doesn't fit the use cases, update the standard.

**Comment by Sunderlandkyl** [comment]
I feel like the format of the standard is clearer than the one that was being used, so I changed the responses for both commands to be closer to that.

I edited the description above to include the updated formats.

I also added a flag to to the vtkPlusCommandRTSCommandResponse class (UseDefaultFormat) that allows responses to be sent with formats other than: 
```
<CommandReply Name="" Status="" Error="" Message=""/>
```

Is there a reason that replies were restricted to the above format?

adamrankin:
Using CommandReply is redundant as the message type is RTSCommand, but it doesn't hurt.

The standard seems to allow for more fine-grained responses/requests.

Are you at NAMIC week? I would confer with the group to come to a consensus.

adamrankin:
So, it looks like the standard wants the command name to be "Get" and we have it designed for specific commands "GetUsParameters" etc...

However, you are specifying individual parameters as part of the command, which means that "Get" should be used.

Do we want to go with a more generic "Get" command that looks up parameters by context?

@lassoan @Sunderlandkyl

lassoan:
It makes things simpler and more robust if we use meaningful command names. A command name such as "Get" is so generic that it is completely useless. You need to inspect the content of the command to decide what to do with it. In contrast, using GetUsParameters as command name allows us to instantiate a vtkPlusUsParameterCommand command class and let it deal with execution of Get/SetUsParameters and related commands.

Sunderlandkyl:
If we're keeping the GetUsParameters/SetUsParameters names, are there any changes that should be made here before it's merged?

Issue #374, opened 2018-06-28
PlusServer not sending correct IGTL header version


adamrankin:
See discussion https://github.com/IGSIO/OpenIGTLinkIO/issues/58

PR #375, opened 2018-06-28
Let PLUS send OpenIGTLinkIO ultrasound meta data

olevs:
This won't work yet, as PLUS only sends OpenIGTLink header version 1. 
Bug report: https://github.com/PlusToolkit/PlusLib/issues/374

After this bug is fixed PLUS should send OpenIGTLinkIO ultrasound meta information for the BK ultrasound scanner.


adamrankin:
This looks like testing code? Was this PR not yet finished?

adamrankin:
I can confirm that Plus sends v2 image messages to clients.

olevs:
Yes, of course you are right. The PR isn't finished. This solution will still require the variables to be defined in the config file. I'll see if I can fix it now.

olevs:
I hope it is better now. Now all PlusTrackedFrame::CustomFrameFields are sent as meta data. 

This way it is simple to extend in the future, by just adding parameters to the FrameFields, like I do in vtkPlusBkProFocusOemVideoSource::AddParametersToFrameFields(). Another solution could be to create a stringlist of the parameters we want to send, but then this list also would have to be maintained.

PR #378, opened 2018-07-07
Updated IntelRealSense device

markasselin:
Main improvements are:

- More efficient code
- Made device more configurable
- Cleaned up code base
- Merged the various RealSense devices from the past few years into one single device


markasselin:
@nehk018 Hi Jorge, I've done some improvements to the IntelRealSense device, would you mind trying them and verifying they achieve what you want? This [sample config](https://github.com/markasselin/PlusLibData/blob/master/ConfigFiles/PlusDeviceSet_Server_IntelRealSenseTracker.xml) should give the identical output to the code you wrote.

Mainly I have:
-Removed the old deprecated "pxc" RealSense device
-Made the device configurable using a config file
-Made the device more efficient computationally (the original implementation took 100% CPU and locked my machine up)
-Updated the documentation.

Please let me know if you have any comments!

PR #381, opened 2018-07-11
Add a device name to KeepAlive status messages

Sunderlandkyl:
@lassoan 

I found that the reason so many status nodes were being created was that the keep alive message had no device name.

In SlicerOpenIGTLink/OpenIGTLinkIF, device name was used for the status node name, so GetFirstNode would not find the existing nodes.

As far as I can see, there are two solutions to this:
1. Add device name "KeepAlive" to KeepAlive status messages
1. Implement some sort of check in OpenIGTLinkIF to avoid creating multiple nodes for messages with no device name. (I imagine this would apply to all types of messges, not just status messages).

Which one (or both) of the solutions should be the preferred approach?

### DIFF
--- a/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
+++ b/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
@@ -1027,6 +1027,7 @@ void vtkPlusOpenIGTLinkServer::KeepAlive()
     for (std::list<ClientData>::iterator clientIterator = this->IgtlClients.begin(); clientIterator != this->IgtlClients.end(); ++clientIterator)
     {
       igtl::StatusMessage::Pointer replyMsg = igtl::StatusMessage::New();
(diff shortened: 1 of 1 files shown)


lassoan:
Nice catch.

For me it looks like a reasonable approach to set an empty device name for messages that we don't want to convert to a node in the scene. If we send a STATUS message with non-empty device name then I would expect that a node with that name will be created in the scene.

So, I would keep the current empty device name in the message and modify the receiver side to ignore STATUS messages with empty device name.

Sunderlandkyl:
Makes sense to me. How does this apply to other message types such as images, that do not have a device name?

Based on the implementation, I would expect that the same thing would happen in that case (many nodes would be created).

lassoan:
Probably we should be consistent and drop messages with empty device name. The only thing that I would do differently for non-STATUS messages is to log a warning when we drop the message.

Issue #169, opened 2017-09-21
Add support for compressed video transfer through OpenIGTLink


Sunderlandkyl:
Imaging speed could be improved by using video compression methods.
Some support has been already added in OpenIGTLink and OpenIGTLinkIO.

https://na-mic.org/wiki/Project_Week_25/Tracked-Ultrasound-Standardization-IV

Migrated from https://app.assembla.com/spaces/plus/tickets/1228/details

-----

TODO:
- [x] Add ability to specify compression level in the config file

PR #384, opened 2018-07-17
re #169: Add option to enable video streaming for image messages

Sunderlandkyl:
When OpenIGTLink is compiled with OpenIGTLink_ENABLE_VIDEOSTREAMING and one of the potential video codecs enabled (Currently just VP9 and H264).

Users can specify the encoding type in the config file to send the image as a compressed VideoMessage.
Transforms are not applied to VideoMessage.

Ex.
```
  <PlusOpenIGTLinkServer 
    MaxNumberOfIgtlMessagesToSend="1" 
    MaxTimeSpentWithProcessingMs="50" 
    ListeningPort="18944" 
    SendValidTransformsOnly="true" 
    OutputChannelId="VideoStream" > 
    <DefaultClientInfo> 
      <MessageTypes> 
        <Message Type="IMAGE" />
      </MessageTypes>
      <ImageNames>
        <Image Name="Image" EmbeddedTransformToFrame="Reference" EncodingType="VP90"/>
      </ImageNames>
    </DefaultClientInfo>
  </PlusOpenIGTLinkServer>
```

See list at re #169 for other TODO.

### DIFF
--- a/src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
+++ b/src/PlusOpenIGTLink/PlusIgtlClientInfo.cxx
@@ -141,9 +141,13 @@ PlusStatus PlusIgtlClientInfo::SetClientInfoFromXmlData(vtkXMLDataElement* xmlda
         continue;
       }
 
+      std::string encodingType;
+      XML_READ_STRING_ATTRIBUTE_NONMEMBER_OPTIONAL(EncodingType, encodingType, imageElem);
+
       ImageStream stream;
       stream.EmbeddedTransformToFrame = embeddedTransformToFrame;
       stream.Name = name;
+      stream.EncodingType = encodingType;
       clientInfo.ImageStreams.push_back(stream);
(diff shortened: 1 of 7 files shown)


Sunderlandkyl:
@lassoan The embedded transform is not applied to the VideoMessage, specifically because there is no embedded  matrix in igtl::VideoMessage (https://github.com/openigtlink/OpenIGTLink/blob/master/Source/VideoStreaming/igtlVideoMessage.h).

lassoan:
For sake of consistency, it would be better if spacing, position, orientation information could be included message. Could you add an issue in OpenIGTLinkIO repository or send an email around to discuss this? If the message is still new then we can probably still change the specification, but we could also decide to save this information into metadata fields.

Sunderlandkyl:
When the pull requests in OpenIGTLink and OpenIGTLinkIO are merged, then this should now send an embedded transform along with the video message.

PR #391, opened 2018-07-27
Add config attribute for saving the currently selected tab in PlusServerLauncher

Sunderlandkyl:
See https://github.com/PlusToolkit/PlusApp/pull/51

### DIFF
--- a/src/PlusCommon/vtkPlusConfig.cxx
+++ b/src/PlusCommon/vtkPlusConfig.cxx
@@ -417,6 +417,19 @@ PlusStatus vtkPlusConfig::LoadApplicationConfiguration()
     saveNeeded = true;
   }
 
+  // Read current tab
+  int currentTab = 0;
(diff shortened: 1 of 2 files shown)


src/PlusCommon/vtkPlusConfig.cxx
```diff
@@ -417,6 +417,19 @@ PlusStatus vtkPlusConfig::LoadApplicationConfiguration()
     saveNeeded = true;
   }
 
+  // Read current tab
+  int currentTab = 0;
```
adamrankin: Considering vtkPlusConfig is used for all Plus applications, and not those with a tabbed gui, perhaps this should be read/written a PlusServerLauncher::ReadConfiguration function? The data can still be saved in PlusConfig.xml

Sunderlandkyl: Good point. I'll add this to PlusServerLauncher then.

PR #409, opened 2018-09-01
Add logarithmic scaling of intensity for WinProbe

dzenanz:


jamesobutler:
Tested with WinProbe hardware and it works. Thanks!

@adamrankin This should be good to go unless you have any comments.

PR #411, opened 2018-09-07
Prevent compression of metaimage when using recording command

### LINKED ISSUE #402: VirtualCapture compresses metaimage with warning saying it can't
When recording with compression enabled and an output file with MHA or MHD extension type I get a warning message about it not being supported.  When I look at the header information of each file types afterwards, it does indicated `CompressedData = True` and by looking at the file size it appears to be compressed.

https://github.com/PlusToolkit/PlusLib/blob/ecfe79335fd2e73d1e78235ec069870fa8dc21d3/src/PlusDataCollection/VirtualDevices/vtkPlusVirtualCapture.cxx#L183-L190

jamesobutler:
This closes #402.

In StartStopRecordingCommand, moving `SetEnableFileCompression` before `OpenFile` prevents the loophole of allowing compressed metaimage files(that are then unable to be uncompressed) when using the StartStopRecordingCommand.

I've also updated the log warning message in `OpenFile` to use the general "metaimage" term since the user could be requesting to record as mha _or_ mhd.


jamesobutler:
@adamrankin Is there anything else needed to move this PR forward? Thanks!

jamesobutler:
_ping_ @adamrankin @lassoan

PR #417, opened 2018-09-17
Register MicronTracker stereo camera as a video source

auneri:
Very basic support that allows streaming to a client (e.g., 3D Slicer) through OpenIGTLink.

### DIFF
--- a/src/PlusDataCollection/MicronTracking/vtkPlusMicronTracker.cxx
+++ b/src/PlusDataCollection/MicronTracking/vtkPlusMicronTracker.cxx
@@ -59,6 +59,12 @@ vtkPlusMicronTracker::vtkPlusMicronTracker()
   this->StartThreadForInternalUpdates = true;
   this->AcquisitionRate = 20;
 
+  this->FrameLeft = vtkSmartPointer<vtkImageData>::New();
+  this->FrameRight = vtkSmartPointer<vtkImageData>::New();
+  this->FrameSize[0] = 0;
+  this->FrameSize[1] = 0;
+  this->FrameSize[2] = 1;
+
   this->IniFile = "MicronTracker.ini";
 }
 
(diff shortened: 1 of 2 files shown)


auneri:
Oh and I also had to remove copying of nonexistent [digiclops](https://github.com/PlusToolkit/PlusLib/blob/Plus-2.6/src/PlusDataCollection/CMakeLists.txt#L402) and [triclops](https://github.com/PlusToolkit/PlusLib/blob/Plus-2.6/src/PlusDataCollection/CMakeLists.txt#L408) libraries for the version of MTC (v3.7.6.8) I have.

adamrankin:
@lassoan what's the precedent for tracker devices that also produce video?

lassoan:
It should be no problem to have a device that can produce both tracking and imaging data. It would probably make sense to have the tracking data, left eye video (and right eye video, if needed) in separate channels, so that tracking data can be interpolated and fused with a different image stream.

src/PlusDataCollection/MicronTracking/vtkPlusMicronTracker.cxx
```diff
@@ -310,6 +335,8 @@ PlusStatus vtkPlusMicronTracker::GetImage(vtkImageData* leftImage, vtkImageData*
 
   int imageWidth = this->MT->mtGetXResolution(-1);
```
adamrankin: Can this function ever return < 0?

src/PlusDataCollection/MicronTracking/vtkPlusMicronTracker.cxx
```diff
@@ -240,6 +246,25 @@ PlusStatus vtkPlusMicronTracker::InternalUpdate()
 #endif
   }
 
+  this->GetImage(this->FrameLeft, this->FrameRight);
+  vtkPlusDataSource* aSource(NULL);
+  for (int i = 0; i < this->GetNumberOfVideoSources(); ++i)
+  {
+    if (this->GetVideoSourceByIndex(i, aSource) != PLUS_SUCCESS)
```
adamrankin: Please add a NotifyConfigure overload that checks that the user has created 2 and exactly 2 video sources for image output.

adamrankin:
Closing as #442 accomplishes the same goal

PR #419, opened 2018-09-27
Reviving PreTMatrix attribute

dzenanz:
This is useful when multiple transforms need to be recorded which differ only by a constant offset.

### DIFF
--- a/src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
+++ b/src/PlusDataCollection/USDigitalEncodersTracking/vtkPlusUSDigitalEncodersTracker.cxx
@@ -86,6 +86,7 @@ public:
   vtkVector3d LocalAxis;
   vtkVector3d LocalAxis2;
 public:
+  vtkSmartPointer<vtkMatrix4x4> BaseToEncoderMatrix = vtkSmartPointer<vtkMatrix4x4>::New();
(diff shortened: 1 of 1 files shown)


dzenanz:
@jamesobutler Can you test the updated code?

jamesobutler:
Tested with the updates and it works. Thanks!

dzenanz:
@lassoan @adamrankin Merge?

lassoan:
PreT is not a valid transform name. Please specify a name that unambiguously tells what coordinate systems it transforms between. For example, BaseToEncoder.

Let us k ow if there is a strong reason to make an exception to the standard naming convention.

jamesobutler:
@lassoan Dzenan issued an update that updated "PreTMatrix" to "BaseToEncoderMatrix".

lassoan:
I missed that, sorry. I just made up BaseToEncoder name as an example. Is that the actual meaning of the transform?

dzenanz:
There is one tracker and multiple transducers. These transducers will all have fixed and known offset relative to tracker. However the geometry might change over time, and we want each transducer's position to be written into the sequence file. Plus does not write constant transforms into sequence file so we "fake" multiple encoders by using this extra translation transform.

But this transform might have some other uses, so `TransducerToEncoder` would be misleading as a name. `BaseToEncoder` would be general enough and expresses the intended use to some extent.

lassoan:
Thank you, I'm merging this now.

I've just found there is a variable "TransformationMatrix" just one line above - which is probably one of the least informative transformation matrix name possible. I would recommend to review coordinate system and transformation names throughout this class.

PR #422, opened 2018-10-02
Minor cleanup and AStyle formatting of USDigital encoders

dzenanz:


lassoan:
Thanks for the update. CurrentTransform name is already better than TransformationMatrix, but to really make it meaningful, you would need to give it a <a>To<b> name, such as ToolToEncoder.

I can merge the pull request as is, or wait for finalization of the name - just let me know how would you like to proceed.

dzenanz:
Are the updated names better? You can merge now.

@jamesobutler @RMR54 This commits updates the name of the property in configuration file to `ToolToEncoderMatrix`. Config files will need to be updated.

jamesobutler:
Ok, no problem. I can update my config files.

lassoan:
Thanks for the updates. I've merged it now.

PR #426, opened 2018-10-10
Reducing usage of std::pair for more explicitly named struct members

adamrankin:
Adding small delay and check of proper server startup (server port not available error not caught)

### DIFF
--- a/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
+++ b/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
@@ -39,6 +39,10 @@ See License.txt for details.
 // OpenIGTLinkIO includes
 #include <igtlioPolyDataConverter.h>
 
+// STL includes
+#include <chrono>
+#include <thread>
+
 #if defined(WIN32)
   #include "vtkPlusOpenIGTLinkServerWin32.cxx"
 #elif defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
```diff
@@ -47,22 +51,25 @@ See License.txt for details.
   #include "vtkPlusOpenIGTLinkServerLinux.cxx"
 #endif
 
-static const double DELAY_ON_SENDING_ERROR_SEC = 0.02;
-static const double DELAY_ON_NO_NEW_FRAMES_SEC = 0.005;
-static const int NUMBER_OF_RECENT_COMMAND_IDS_STORED = 10;
-static const int IGTL_EMPTY_DATA_SIZE = -1;
-
-const float vtkPlusOpenIGTLinkServer::CLIENT_SOCKET_TIMEOUT_SEC = 0.5;
+namespace
```
lassoan: If you use anonymous namespace then there is no need for static to make the variables available only in the current translation unit. You can remove the static keywords.

src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
```diff
@@ -155,16 +160,24 @@ PlusStatus vtkPlusOpenIGTLinkServer::StartOpenIGTLinkService()
...
 
   if (this->DataSenderThreadId < 0)
   {
-    this->DataSenderActive.first = true;
+    this->DataSenderActive.Request = true;
     this->DataSenderThreadId = this->Threader->SpawnThread((vtkThreadFunctionType)&DataSenderThread, this);
   }
 
+  // Wait a short duration to see if both threads initialized properly
+  std::this_thread::sleep_for(std::chrono::milliseconds(250));
```
lassoan: self->ServerSocket->CreateServer(self->ListeningPort) is a networking task, so we cannot be sure that it completes within 250ms. The nicest would be to wait (with a timeout) for a mutex to be released, but retrying 10x-20x with some small delay between them would be a bit faster and more robust.

src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
adamrankin: I thought so, I wasn't sure. Done.

src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
adamrankin: What timeout would you recommend? I didn't want to delay server startup by very much, as this use case occurs very infrequently.

lassoan: You can use shorter timeouts (50ms) with multiple repeated checks (if not ready yet then wait again and re-check). Maybe up to 2 seconds in total.

PR #428, opened 2018-10-19
Adding argument parsing to script to enable --track argument to CTest

adamrankin:
@lassoan would you check that the changes to individual test running can be easily updated in any scripts you run?


adamrankin:
When we integrate this PR, I will make similar commits to PlusApp and PlusBuild

Issue #313, opened 2018-01-09
Tool status is simplified from ToolStatus to bool when sent via PlusServer


adamrankin:
If possible, update the channel/buffer/trackedframelist to store more complete tool status information and send this via metadata attached to Transform/TDATA/etc.. messages.

adamrankin:
See vtkPlusTransformRepository::TransformInfo and vtkPlusTransformRepository::GetTransformValid

PR #432, opened 2018-10-28
Addressing #313 ToolStatus simplified to bool

adamrankin:
Refactoring code to use ToolStatus throughout to represent the current state of a transform.

Once accepted, PlusApp commit to match and update SlicerOpenIGTLink if desired to store metadata status as node attribute


adamrankin:
Confirmed working. See https://github.com/openigtlink/SlicerOpenIGTLink/pull/65 for matching Slicer side.

Sunderlandkyl:
@adamrankin Is there a corresponding PlusApp commit for this?

Issue #383, opened 2018-07-17
Update OptiTrack device to use Motive 2.0


markasselin:
After merging these changes Plus2.7 nightlies will require OptiTrack Motive v2.0, and the stable (Plus 2.6) will continue to support the older versions of Motive (v1.10.3 recommended).

markasselin:
Completed

PR #438, opened 2018-11-07
re #383: Updated OptiTrack device to use Motive 2.0

markasselin:


adamrankin:
Perfect, sounds good.

PR #449, opened 2018-11-23
Cleanup of MicronTrackerInterface

### LINKED ISSUE #95: Clean up MicronTracker interface classes
MicronTracker tracking seems to work fine, but the interface classes (PlusLib\src\Tracking\MicronTracking\Utils) are implemented very poorly. Need a review and clean-up including:
VERY IMPORTANT PROGRAMMING NOTE in MicronTrackerInterface.cpp
check ownedByMe handling (probably it should be set to false if the handle is not created by new)
consistent return values (sometimes int, sometimes bool; sometimes non-zero means error, sometimes 0, false)
check timestamp computation (mtGetLatestFrameTime(cam) function can provide accurate enough without filtering)

There also seem to be a larger time lag in the tracking in Plus than in the MicronTracker CPP demo.

Migrated from https://app.assembla.com/spaces/plus/tickets/realtime_list?page=1&ticket=571

adamrankin:
Consistent return value of bool for success/failure condition

Code cleanup and standardization

Resolves #95


src/Utilities/MicronTrackerInterface/Cameras.cpp
```diff
@@ -11,89 +11,83 @@
...
+  Cameras_HistogramEqualizeImagesGet(&R);
+  return R;
 }
 
-/****************************/
-int Cameras::setHistogramEqualizeImages(bool on_off)
+//------------------------------------------
+bool Cameras::setHistogramEqualizeImages(bool on_off)
 {
-	int result = Cameras_HistogramEqualizeImagesSet(on_off);
```
lassoan: This was not a Boolean return value but an error code. By changing it to Boolean we don't know what was the exact failure.

src/Utilities/MicronTrackerInterface/Cameras.cpp
```diff
@@ -11,89 +11,83 @@
...
 Cameras::~Cameras()
 {
   // Clear all previously connected camera
-  for (std::vector<MCamera *>::iterator camsIterator = m_vCameras.begin(); camsIterator != m_vCameras.end(); ++camsIterator)
+  for (std::vector<MCamera*>::iterator camsIterator = CameraList.begin(); camsIterator != CameraList.end(); ++camsIterator)
   {
     free(*camsIterator);
   }
-  m_vCameras.clear();
+  CameraList.clear();
```
lassoan: If you remove m_ prefix then please add this-> to make it clear that it is a member variable.

src/Utilities/MicronTrackerInterface/Cameras.cpp
adamrankin: Well, that's a pattern over the whole API. The int return type was actually the #define's in MTC.h

If you want to keep that API behaviour, I can wrap those defines in an enum and return that instead.

adamrankin:
>My only concern is the suppressing of error codes.
I agree. See reply to inline comment.

src/Utilities/MicronTrackerInterface/Cameras.cpp
adamrankin: Will do.

src/Utilities/MicronTrackerInterface/Cameras.cpp
lassoan: My point is that there is no need to leak out MTC.h through the wrapper's interface if there are methods the wrapper that give info about the error code.

If MTC.h only contains these error codes then it may be fine to make it available outside, but preferably the wrapper should hide all internals.

adamrankin: hmm, no, my solution only works with c++11

adamrankin:
@lassoan after much research into c++ language features, it seems there's no clean way of wrapping a C enum without exposing anything else in the header include.

Considering this library would only need changing if a new MTC version changed the enums, it seems like a reasonable enough approach to redefine them as a new interface enum.

lassoan:
No need to expose the enums, it may be enough to expose a method that can convert from error code to human-readable error message. If there are only a handful of error codes then redefining enum on the interface makes sense, too.

PR #452, opened 2018-11-27
Spatial compounding

dzenanz:
@RMR54 @tomekcz and @jamesobutler are testing this.


src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1498,16 +1498,21 @@ IF(PLUS_USE_WINPROBE_VIDEO)
 
   LIST(APPEND ${PROJECT_NAME}_LIBS
     ${WINPROBESDK_DIR}/${Bitness}/$<$<CONFIG:Debug>:Debug>$<$<NOT:$<CONFIG:Debug>>:Release>/UltraVisionManagedDll${CMAKE_STATIC_LIBRARY_SUFFIX}
```
adamrankin: Does $<CONFIG:Debug> and similar work for multi config generators like xcode or msvc?

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1498,16 +1498,21 @@ IF(PLUS_USE_WINPROBE_VIDEO)
...
     ${WINPROBESDK_DIR}/${Bitness}/Release/UltraVisionWrapper${CMAKE_SHARED_LIBRARY_SUFFIX}
+    ${WINPROBESDK_DIR}/${Bitness}/Release/UltraVision.DXWrapper${CMAKE_SHARED_LIBRARY_SUFFIX}
     ${WinProbeAssemblies}
     ${WinProbeXMLFiles}
     )
   LIST(APPEND External_Libraries_Debug
     ${WINPROBESDK_DIR}/${Bitness}/Debug/UltraVisionManagedDll${CMAKE_SHARED_LIBRARY_SUFFIX}
+    ${WINPROBESDK_DIR}/${Bitness}/Debug/UltraVisionManagedDXDll${CMAKE_SHARED_LIBRARY_SUFFIX}
     ${WINPROBESDK_DIR}/${Bitness}/Debug/UltraVisionWrapper${CMAKE_SHARED_LIBRARY_SUFFIX}
+    ${WINPROBESDK_DIR}/${Bitness}/Release/UltraVision.DXWrapper${CMAKE_SHARED_LIBRARY_SUFFIX}
```
adamrankin: Is this a typo or is there only a release version of this library?

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
```diff
@@ -25,148 +25,165 @@
...
 
-    /* Set the transmit frequency of US probe (MHz) */
-    PlusStatus SetTxTxFrequency(float frequency);
+  /* Set the transmit frequency of US probe (MHz) */
+  PlusStatus SetTxTxFrequency(float frequency);
 
-    /* Get the transmit frequency of US probe (MHz) */
-    float GetTxTxFrequency();
+  /* Get the transmit frequency of US probe (MHz) */
+  float GetTxTxFrequency();
```
adamrankin: Duplicate on purpose?

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
```diff
@@ -25,148 +25,165 @@
...
 
-    /* Set the scan depth of US probe (mm) */
-    PlusStatus SetSSDepth(float depth);
+  /* Set the scan depth of US probe (mm) */
+  PlusStatus SetSSDepth(float depth);
 
-    /* Get the scan depth of US probe (mm) */
-    float GetSSDepth();
+  /* Get the scan depth of US probe (mm) */
+  float GetSSDepth();
```
adamrankin: No need to save characters, easier for autocomplete tools if the user could type "scan" and this method would show up.

Recommend GetScanDepth unless you think otherwise.

adamrankin: And as per @lassoan suggestion, add unit to end of function name

src/PlusDataCollection/CMakeLists.txt
dzenanz: These wrappers are compiled primarily (or even only) using MSVC. I expect that this would work for XCode too.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -16,526 +16,700 @@ See License.txt for details.
...
+//-----------------------------------------------------------------------------
+void vtkPlusWinProbeVideoSource::ReconstructFrame(char* data)
 {
-    thisPtr->FrameCallback(length, data, hHeader, hGeometry);
-    return length;
+  uint16_t* frame = reinterpret_cast<uint16_t*>(data + 16);
+  assert(m_BModeBuffer.size() == m_samplesPerLine * m_transducerCount);
+  const float logFactor = m_OutputKnee / std::log(1 + m_Knee);
+
+  #pragma omp parallel for
```
adamrankin: I don't see any OpenMP checking in the CMake. Will this compile if openmp is not available?

src/PlusDataCollection/CMakeLists.txt
dzenanz: Typo 😠 (angry at myself). Thanks for pointing it out!

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
dzenanz: Short answer: yes.

Long answer: WinProbeSDK exposes a lot of internal parameters, and uses prefixes for groups of parameters to avoid name collisions. E.g. `BMode::IsEnabled` gets exposed as `GetBIsEnabled` and `SpatialCompound::IsEnabled` gets exposed as `GetSCIsEnabled`. So `Transmit` class gets prefix `Tx`, and property `TxFrequency` gets exposed as `GetTxTxFrequency`.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.h
dzenanz: OK I will systematically review and change method and ivar names.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
dzenanz: Unrecognized `#pragma`s should be ignored by the compiler. See https://stackoverflow.com/a/17511382/276168

PR #458, opened 2018-12-03
Migrate PlusCommon classes to IGSIO

Sunderlandkyl:
See https://github.com/IGSIO/IGSIO/pull/2


adamrankin:
Doing a test build, if it passes, will merge.

PR #469, opened 2018-12-09
Allow the fps of MMFVideo devices to be reduced to the desired acquis…

Sunderlandkyl:
…ition rate by dropping extra frames

- When a frame-rate that is unsupported by a webcam, frames that are acquired faster than the acquisition rate are dropped
- This allows applications to reduce bandwidth to levels that are not supported by the webcam

### DIFF
--- a/src/PlusDataCollection/MicrosoftMediaFoundation/vtkPlusMmfVideoSource.cxx
+++ b/src/PlusDataCollection/MicrosoftMediaFoundation/vtkPlusMmfVideoSource.cxx
@@ -668,7 +668,29 @@ PlusStatus vtkPlusMmfVideoSource::AddFrame(unsigned char* bufferData, DWORD buff
     LOG_ERROR("Unable to retrieve the video source in the media foundation capture device.");
     return PLUS_FAIL;
   }
-  PlusStatus status = aSource->AddItem(&this->UncompressedVideoFrame, this->FrameIndex);
+
+  const double maximumFrameTimeVariance = 0.2;  // to make sure we don't drop frames because of slight variance in acquisition rate, we allow up to 20% higher frame rate before we start dropping frames
+  double acquisitionRate = this->GetAcquisitionRate();
+  double minimumTimeBetweenBetweenRecordedFramesSec = (1.0 - maximumFrameTimeVariance) / this->GetAcquisitionRate();
+  double lastFrameTimeSec = -1.0;
+  double currentTime = vtkIGSIOAccurateTimer::GetSystemTime();
+
+  StreamBufferItem latestFrame;
+  if (aSource->GetNumberOfItems() > 2 && aSource->GetLatestStreamBufferItem(&latestFrame) == ITEM_OK)
+  {
+    lastFrameTimeSec = latestFrame.GetUnfilteredTimestamp(0.0);
+    double secondsSinceLastFrame = currentTime - lastFrameTimeSec;
+    if (lastFrameTimeSec > 0 && secondsSinceLastFrame < minimumTimeBetweenBetweenRecordedFramesSec)
+    {
+      // For some webcams, the requested acquistion rate may not be availiable (not supported, or error in configuration).
+      // In this case we can artificially limit frames to the requested acquisition rate by ignoring frames.
+
+      // The required time has not elapsed between frames.
+      // Do not need to record this frame.
+      return PLUS_SUCCESS;
+    }
+  }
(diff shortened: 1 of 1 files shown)


src/PlusDataCollection/MicrosoftMediaFoundation/vtkPlusMmfVideoSource.cxx
```diff
@@ -609,6 +611,16 @@ PlusStatus vtkPlusMmfVideoSource::AddFrame(unsigned char* bufferData, DWORD buff
     return PLUS_SUCCESS;
   }
 
+  double targetSecondsBetweenFrames = 1.0 / this->GetAcquisitionRate();
+  double secondsSinceLastFrame = vtkIGSIOAccurateTimer::GetSystemTime() - this->LastFrameTime;
+  if (this->LastFrameTime > 0 && secondsSinceLastFrame < targetSecondsBetweenFrames)
+  {
+    // The required time has not elapsed between frames.
```
lassoan: explain why this can happen for this video source (i.e., you don't always get what you request)

src/PlusDataCollection/MicrosoftMediaFoundation/vtkPlusMmfVideoSource.cxx
```diff
@@ -614,6 +614,26 @@ PlusStatus vtkPlusMmfVideoSource::AddFrame(unsigned char* bufferData, DWORD buff
   {
     return PLUS_FAIL;
   }
+
+  const double maximumFrameTimeVariance = 0.2;  // to make sure we don't drop frames because of slight variance in acquisition rate, we allow up to 20% higher frame rate before we start dropping frames
```
adamrankin: This will not work as `this->AcquisitionRate` is passed in `InternalConnect` as the desired webcam framerate.

You will need to add another variable in the XML to specify the throttled framerate.

adamrankin: Or repurpose `this->AcquisitionRate` to mean throttled framerate and add another XML attribute like `WebcamFramerate` and update the docs/sample configs.

lassoan: MMF imaging devices support a set of frame rates predefined at the driver level and you may ask for 5fps but the closest you may get is 30fps.

this->AcquisitionRate means the desired (throttled, if necessary) frame rate.

adamrankin: If you ask for 5fps and the device doesn't support it, the request will fail. The only exception is I added double rounding so if you ask for 60fps and the device supports 59.99999999 it will allow it.

adamrankin: The MediaFoundationLibrary does not return closest match. We can change it to do that if we want to.

adamrankin: Something isn't right, some of my edits are not in the MfCapture lib. Investigating.

Sunderlandkyl:
Alright, this behaves as expected now.
Had some issues with comparing videoSource->GetLatestTimestamp() (Filtered) and vtkIGSIOAccurateTimer::GetSystemTime() (Unfiltered), but that should be resolved.

adamrankin:
Just to confirm, you've tested when requested == throttled == webcam and it behaves as expected?

Sunderlandkyl:
Yes, doesn't drop any frames if the requested framerate is the same as the actual framerate.

jamesobutler:
Does it make sense (for maintenance reasons) to encourage the use of the mmf implementation within the [OpenCVVideo](http://perk-software.cs.queensu.ca/plus/doc/nightly/user/DeviceOpenCVVideo.html) device? I know the origin of the code used here in Plus is the same that was put into OpenCV though it has been slightly updated over there.

adamrankin:
If CUDA is not enabled in OpenCV, it's not that bad of a dependency. The build on my machine takes a couple of minutes. I don't see it as that bad of a dependency, and have OpenCV video device enabled on all my dev builds.

I would have to examine the OpenCV code to see how they use the MfVideoCapture API.

jamesobutler:
Just thought I throw that out there. I figured reducing development on the same mmf is probably the right thing to do. There’s been a good number of commits to mmf work in openCV that I wondered how that same work could easily be brought over to Plus.

PR #487, opened 2019-01-23
Fix SendText command in vtkPlusGenericSerialDevice

Sunderlandkyl:
Due to a change in the signature of SendText() in vtkPlusGenericSerialDevice, the method was no longer overridden.
This commit restores the original function signature by creating a new function with the original signature that calls the modified function.


lassoan:
Good catch!

PR #494, opened 2019-02-11
Added support for PointGrey Spinnaker API compatible devices

markasselin:


src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1645,6 +1646,107 @@ IF(PLUS_USE_INTERSONARRAYSDKCXX_VIDEO)
     )
 ENDIF()
 
+# --------------------------------------------------------------------------
+# Point Grey Spinnaker API support
+
+IF(PLUS_USE_SPINNAKER_VIDEO)
+  message("ABCD")
```
adamrankin: leftover debugging

adamrankin: I forget, is this where we typically put `find_package(Spinnaker...)`?

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1645,6 +1646,107 @@ IF(PLUS_USE_INTERSONARRAYSDKCXX_VIDEO)
...
+  # Collect all the Spinnaker API libraries
+  LIST(APPEND ${PROJECT_NAME}_INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/PointGrey
+    ${SPINNAKER_API_INCLUDE_DIR}
+    )
+  
+  IF(WIN32)
+    # Windows libs
+    LIST(APPEND ${PROJECT_NAME}_LIBS
+      debug     ${SPINNAKER_API_LIBRARY_DIR}/Spinnakerd_v140${CMAKE_STATIC_LIBRARY_SUFFIX}
```
adamrankin: oohhhh nice!

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -1645,6 +1646,107 @@ IF(PLUS_USE_INTERSONARRAYSDKCXX_VIDEO)
...
+      ${SPINNAKER_API_BINARY_DIR}/log4cpp.spinnaker.property
+      ${SPINNAKER_API_BINARY_DIR}/log4cpp_MD_VC140_v3_0${CMAKE_SHARED_LIBRARY_SUFFIX}
+      ${SPINNAKER_API_BINARY_DIR}/MathParser_MD_VC140_v3_0${CMAKE_SHARED_LIBRARY_SUFFIX}
+      ${SPINNAKER_API_BINARY_DIR}/NodeMapData_MD_VC140_v3_0${CMAKE_SHARED_LIBRARY_SUFFIX}
+      ${SPINNAKER_API_BINARY_DIR}/Spinnaker_v140${CMAKE_SHARED_LIBRARY_SUFFIX}
+      ${SPINNAKER_API_BINARY_DIR}/XMLParser_MD_VC140_v3_0${CMAKE_SHARED_LIBRARY_SUFFIX}
+      )
+
+  ELSEIF(UNIX)
+    # Ubuntu libs
```
adamrankin: this is also OSX, if both are ok, no change, but you can do `ELSEIF(UNIX AND NOT MAC)`

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+
+#include "PlusConfigure.h"
+#include "vtkPlusSpinnakerVideoSource.h"
+
+// Local includes
+#include "vtkPlusChannel.h"
+#include "vtkPlusDataSource.h"
+
+// Spinnaker API includes
+#include "Spinnaker.h"
```
adamrankin: total preference, change not needed, but I like to do external includes with `<...>` instead of `"..."`

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+    this->FrameSize[1] = static_cast<unsigned int>(requestedFrameSize[1]);
+    this->FrameSize[2] = 1;
+  }
+
+  // frame rate
+  XML_READ_SCALAR_ATTRIBUTE_OPTIONAL(int, FrameRate, deviceConfig);
+
+  // exposure
+  XML_READ_ENUM3_ATTRIBUTE_OPTIONAL(ExposureMode, deviceConfig,
+    "TIMED", EXPOSURE_TIMED,
```
adamrankin: If you re-use these strings with PrintConfiguration, I would move them to static/anonymous vars

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+  return this->CheckCameraParameterValidity();
+}
+
+//-----------------------------------------------------------------------------
+PlusStatus vtkPlusSpinnakerVideoSource::WriteConfiguration(vtkXMLDataElement* rootConfigElement)
+{
+  LOG_TRACE("vtkPlusSpinnakerVideoSource::WriteConfiguration(vtkXMLDataElement* rootConfigElement)");
+
+  XML_FIND_DEVICE_ELEMENT_REQUIRED_FOR_WRITING(deviceConfig, rootConfigElement);
+
```
adamrankin: Please complete this function

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+  return PLUS_FAIL;
+}
+
+//----------------------------------------------------------------------------
+PlusStatus vtkPlusSpinnakerVideoSource::InternalConnect()
+{
+  LOG_TRACE("vtkPlusSpinnakerVideoSource::InternalConnect()");
+  
+  // initiailize system pointer
```
adamrankin: typo

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+    {
+      this->Internal->CameraPtr->BalanceWhiteAuto.SetValue(Spinnaker::BalanceWhiteAutoEnums::BalanceWhiteAuto_Once);
+    }
+    else if (this->WhiteBalanceMode == WB_AUTO_CONTINUOUS)
+    {
+      this->Internal->CameraPtr->BalanceWhiteAuto.SetValue(Spinnaker::BalanceWhiteAutoEnums::BalanceWhiteAuto_Continuous);
+    }
+
+    // set sharpening mode && sharpening (if manual sharpening control enabled)
+    //if (this->SharpeningMode == SHARPENING_MANUAL)
```
adamrankin: Please remove if not needed

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+  try
+  {
+    this->Internal->CameraPtr->DeInit();
+    this->Internal->CameraPtr = NULL;
+    this->Internal->CameraList.Clear();
+    this->Internal->SystemPtr->ReleaseInstance();
+  }
+  catch (Spinnaker::Exception &e)
+  {
+    LOG_ERROR("SpinnakerVideoSource: Failed to de-initialize camera. Exception text: " << e.what());
```
adamrankin: This needs to guarantee as best as possible that memory has been deallocated. Could you make the try catch block more granular so you can more accurately detect what failed and compensate in each catch block?

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+    return PLUS_FAIL;
+  }
+
+  return PLUS_SUCCESS;
+}
+
+//----------------------------------------------------------------------------
+PlusStatus vtkPlusSpinnakerVideoSource::NotifyConfigured()
+{
+  // TODO: Implement some configuration checks here
```
adamrankin: Please implement. Some basic checks like number of video sources >= or == 1, number of output channels >= or == 1, etc... would be fine

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+    {
+      LOG_ERROR("Invalid image type selected.");
+      return PLUS_FAIL;
+    }
+
+    // add image to PLUS buffer
+
+    // get buffer
+    vtkPlusDataSource* videoSource(NULL);
+    if (this->GetFirstVideoSource(videoSource) != PLUS_SUCCESS)
```
adamrankin: Depending on how you set up your expectations in NotifyConfigured you could adjust this code (single or multiple data sources, etc...)

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+        videoSource->SetPixelType(VTK_UNSIGNED_CHAR);
+        videoSource->SetNumberOfScalarComponents(1);
+        videoSource->SetInputFrameSize(this->FrameSize);
+      }
+    }
+    
+    // add frame to PLUS buffer
+    if (this->PixelEncoding == RGB24 || this->PixelEncoding == BGR24)
+    {
+      retVal = videoSource->AddItem(
```
adamrankin: Plus pixel encoding is RGB24, may have to convert from BGR24 (had this bug in OpenCV video device)

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.h
```diff
@@ -0,0 +1,152 @@
...
+  PlusStatus InternalStartRecording();
+
+  /*! Stop recording */
+  PlusStatus InternalStopRecording();
+
+  /*! Is this device a tracker */
+  bool IsTracker() const { return false; }
+  bool IsVirtual() const { return false; }
+  
+  /*! Get an update from the tracking system and push the new transforms to the tools. This function is called by the tracker thread.*/
```
adamrankin: remove or fix comment

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.h
```diff
@@ -0,0 +1,152 @@
...
+  /*! Read configuration from xml data */
+  virtual PlusStatus ReadConfiguration(vtkXMLDataElement* config);
+  /*! Write configuration to xml data */
+  PlusStatus WriteConfiguration(vtkXMLDataElement* config);
+
+  /*!
+  Record incoming data at the specified acquisition rate.  The recording
+  continues indefinitely until StopRecording() is called.
+  */
+  PlusStatus InternalStartRecording();
```
adamrankin: InternalXYZ functions need to be in a `protected:` section

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
```diff
@@ -0,0 +1,807 @@
...
+
+//-----------------------------------------------------------------------------
+PlusStatus vtkPlusSpinnakerVideoSource::ReadConfiguration(vtkXMLDataElement* rootConfigElement)
+{
+  LOG_TRACE("vtkPlusSpinnakerVideoSource::ReadConfiguration(vtkXMLDataElement* rootConfigElement)");
+
+  XML_FIND_DEVICE_ELEMENT_REQUIRED_FOR_READING(deviceConfig, rootConfigElement);
+
+  // camera params
+  XML_READ_SCALAR_ATTRIBUTE_REQUIRED(int, CameraNumber, deviceConfig);
```
adamrankin: Would you consider making camera number optional and using the first available camera if not defined?

src/PlusDataCollection/CMakeLists.txt
markasselin: Oops... removed

markasselin: We typically put the find_package() calls in the PlusBuild CMakeLists.txt

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
markasselin: As a std::map?

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
markasselin: Ah okay - it was not clear to me what should go in this method. Perhaps we should write a documentation page about the standard device methods for a subclass of vtkPlusDevice?

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
markasselin: Yes; however, I intended to set this device up so that PLUS could pipe BGR info through in case the consuming code was relying on OpenCV. Otherwise we'd be doing a BGR->RGB conversion only to do the opposite in the consuming code.

adamrankin: I agree that that makes sense, however color in Plus is explicitly RGB. If we want to add pixel ordering options, then we'd need to extend data source configuration and handling. If you want that, please add an issue and we can plan it out.

src/PlusDataCollection/PointGrey/vtkPlusSpinnakerVideoSource.cxx
adamrankin: Nah, just a bunch of static or anonymous namespaced std::string

adamrankin: Or const char* if they're mostly used in VTK functions that don't like std::string

PR #508, opened 2019-03-14
Added support for clarius ultrasound scanner

zoez526:
Added support for [clarius ultrasound scanner](https://support.clarius.com/hc/en-us/articles/360020008631-Listen-API). Please let me know if anything needs to be changed.


adamrankin:
I've made some changes. I'll integrate the pull request and make sure it compiles, but could you test it to make sure it still works?

PR #511, opened 2019-03-19
BUG: application which uses Plus can configure without Qt

dzenanz:
If Plus is configured without Qt, an application which uses Plus fails to configure:

```text
CMake Error at C:/Dev/PlusGit/s64/PlusLib-bin/src/PlusLibConfig.cmake:54 (FIND_PACKAGE):
  Found package configuration file:

    C:/Libs/Qt/5.11.1/msvc2015_64/lib/cmake/Qt5/Qt5Config.cmake

  but it set Qt5_FOUND to FALSE so package "Qt5" is considered to be NOT
  FOUND.  Reason given by package:

  The Qt5 package requires at least one component

Call Stack (most recent call first):
  CMakeLists.txt:14 (FIND_PACKAGE)
```

The generated PlusLibConfig.cmake contains:

```cmake
SET(PLUSLIB_QT_COMPONENTS )
FIND_PACKAGE(Qt5 REQUIRED COMPONENTS )
```


dzenanz:
@sjh26 this should fix your issue.

Sunderlandkyl:
Thanks!

PR #523, opened 2019-04-05
BUG: Update WinProbe internal start and stop recording

RMR54:
This also fixes SetFocalPointDepth() tracebacks observed after freezing/unfreezing probe.

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -692,7 +692,12 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   std::this_thread::sleep_for(std::chrono::milliseconds(100));
   if(m_Mode == Mode::M)
   {
-    this->SetMModeEnabled(true);
+    SetMIsEnabled(true);
+    SetMIsRevolving(m_MRevolvingEnabled);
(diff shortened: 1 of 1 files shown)


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -692,7 +692,12 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   std::this_thread::sleep_for(std::chrono::milliseconds(100));
   if(m_Mode == Mode::M)
   {
-    this->SetMModeEnabled(true);
+    SetMIsEnabled(true);
```
dzenanz: Calling `this->SetMModeEnabled(true)` would avoid code duplication. Please amend this commit.

RMR54: Do I remove the check `if(Connected)` within `setMModeEnabled`? Or, if I moved `this->SetMModeEnabled(true)`; to `InternalStartRecording()` is it okay to apply all M-Mode settings when unfreezing probe?

dzenanz: I forgot that during `InternalConnect`, connected is still `false`. Leave the code as-is.

jamesobutler:
@adamrankin or @lassoan Could you merge this into master for us? Thanks!

PR #546, opened 2019-05-06
BUG: B Mode texture copy wan't using the correct row pitch

sjh26:


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -404,10 +404,10 @@ void vtkPlusWinProbeVideoSource::FrameCallback(int length, char* data, char* hHe
         int slicePitch;
         int rowPitch;
         int tLength = WPDXGetFusedTexData(&texture, &slicePitch, &rowPitch);
-        assert(tLength == frameSize[0] * frameSize[1]);
+        //assert(tLength == frameSize[0] * frameSize[1]);
```
jamesobutler: If this line was uncommented, would it still make sense? Or is this obsolete code that can be removed?

lassoan: Yes, if it is important to check this then add proper error handling (log message, return with error, etc.), if not important then just remove the comment.

sjh26: Right, good catch.  The assert as is no longer makes sense, as the texture returned from the API is no longer guaranteed to have the same size as the image frame.  The updated check would be:  tLength == frameSize[0] * rowPitch.  Will put in proper error handling for this and push another commit.

RMR54: Tested recent assert update with hardware, LGTM

PR #548, opened 2019-05-09
BUG: Include B-mode frames when in BRF mode.

RMR54:
B-mode frames were being ignored when in BRF mode. B-mode frame header (in BRF mode) is now included within `FrameCallback()` checks.

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -306,7 +306,7 @@ void vtkPlusWinProbeVideoSource::FrameCallback(int length, char* data, char* hHe
       AdjustBufferSizes();
       AdjustSpacing();
     }
-    else if(usMode & BFRFALineImage_RFData && frameSize[0] != m_SamplesPerLine * ::GetSSDecimation())
+    else if(usMode & BFRFALineImage_RFData && frameSize[0] != m_SamplesPerLine * m_SSDecimation)
     {
       LOG_INFO("Rf frame size updated. Adjusting buffer size and spacing.");
       AdjustBufferSizes();
@@ -691,6 +691,8 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   this->SetTransmitFrequencyMHz(m_Frequency);
   this->SetVoltage(m_Voltage);
   this->SetScanDepthMm(m_ScanDepth);
+  // Update decimation variable on start, based on scan depth
+  m_SSDecimation = ::GetSSDecimation();
   this->SetSpatialCompoundEnabled(m_SpatialCompoundEnabled); // also takes care of angle  and count
 
   //setup size for DirectX image
@@ -868,6 +870,8 @@ PlusStatus vtkPlusWinProbeVideoSource::SetScanDepthMm(float depth)
     SetPendingRecreateTables(true);
     //what we requested might be only approximately satisfied
     m_ScanDepth = ::GetSSDepth();
+    // Update decimation with scan depth
+    m_SSDecimation = ::GetSSDecimation();
     if(Recording)
     {
       WPExecute();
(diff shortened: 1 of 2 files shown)


RMR54:
@dzenanz, @sjh26 could you please test this as well with the latest Winprobe API updates?

RMR54:
@sjh26 it looks like when setting this up in "BRF" mode and then setting scan depth, it results in atypical stream issue that we've encountered before. This doesn't happen if setting scan depth when in Mode="M"

RMR54:
@sjh26, the branch should now work with latest UV API updates. The atypical stream was due to errors in dlls, the fix for which is included the latest commit of UV API repo. I've tested scan depth updates and I'm not getting any errors.

RMR54:
I've also now exposed setting Rf-mode, thereby allowing switching between M-mode to Rf-mode, the extraSource now displays Rf stream when in `m_Mode = Mode::BRF` , and M-mode frames when in `m_Mode = Mode::M`

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -449,13 +465,13 @@ void vtkPlusWinProbeVideoSource::FrameCallback(int length, char* data, char* hHe
   }
   else if(usMode & BFRFALineImage_RFData)
   {
-    assert(length == frameSize[1] * brfGeometry->Decimation * frameSize[0] * sizeof(int32_t));
-    FrameSizeType frameSizeRF = { frameSize[1]* brfGeometry->Decimation, frameSize[0], 1 }; // x and y axes flipped on purpose
     for(unsigned i = 0; i < m_ExtraSources.size(); i++)
     {
+      assert(length == frameSize[0] * frameSize[0] * sizeof(int32_t));
```
dzenanz: [0] * [0] -> [0] * **[1]** (Index zero repeated twice)

RMR54: Good catch! I've updated it now

dzenanz:
There are some further [fixes](https://github.com/dzenanz/PlusLib/commit/6dcb1125006ed7e660facf76254892d12c653d3b) for switching modes in my fork's [rf-mode-b-mode](https://github.com/dzenanz/PlusLib/tree/rf-mode-b-mode) branch. There is still some crash in WinProbe's code when switching RF->M->RF. I will try to debug it next.

RMR54:
Thanks @dzenanz. I thought we would implement calls outside of class to disable one mode prior to enabling the other - that's how I 've been testing functionality. We could also implement it within this class. It would probably have to be expanded with additional modes added/exposed, i.e. we'd have to disable all other modes prior to enabling one.

dzenanz:
OK I think the last [commit](https://github.com/dzenanz/PlusLib/commit/5962871bbe9d4a2aa3a45dc9a468694d3f0a7431) fixes the mode switching problem. @RMR54  can you test too?

RMR54:
Thanks @dzenanz, looks good with testing. Switching's easier now, with handling disabling of other modes within rf/m-mode enable functions. Would it be prudent to check if GetMModeEnabled() [here](https://github.com/dzenanz/PlusLib/commit/5962871bbe9d4a2aa3a45dc9a468694d3f0a7431#diff-bb24314bd63c803aa01f3777b7ed4a5aR1076) and GetBRFEnabled() [here](https://github.com/dzenanz/PlusLib/blob/5962871bbe9d4a2aa3a45dc9a468694d3f0a7431/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx#L1126)?

dzenanz:
It is more of a precautionary matter, but you could try it.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -1073,6 +1073,11 @@ void vtkPlusWinProbeVideoSource::SetBRFEnabled(bool value)
 {
   if(Connected)
   {
+    if(GetMModeEnabled())
```
dzenanz: Calling this class' setters in both of these setters leads to infinite recursion. Did you try this? Why not just cherry-pick my commit(s)?

RMR54: Yes, you're right @dzenanz, I've cherry-picked your commits and updated this PR. Thanks!

PR #563, opened 2019-06-06
BUG: Update WP RF stream pixel spacing

RMR54:
Pixel spacing has been updated for Rf stream, to account for orientation updates. Decimation factor is also accounted, to get the correct pixel spacing in y (which is in the case for Rf, ElementSpacing[0])


RMR54:
@dzenanz, could you please review this?

PR #567, opened 2019-06-17
BUG: Do not update m_mode when getting M-Mode/Rf mode disabled status

RMR54:
m_Mode was getting to set to B-mode, even if M-mode had been enabled. Update m_Mode now only when mode is found to be enabled in getter functions.


RMR54:
@dzenanz, @sjh26, Could you please review this? I was getting some API tracebacks earlier, after updating transform (calling `GetBRFEnabled()`), with [m_Mode](url) being incorrectly updated.

PR #572, opened 2019-06-26
ENH: Add new buffer call, with WinProbe API updates for frame processing

RMR54:
Set buffer max to 1, and match API updates that optimized frame processing.


RMR54:
Although I didn't observe errors or tracebacks, I just got news that `WPDXSetFusedTexBufferMax(1); ` should be called after `WPDXSetIsGetSpatialCompoundedTexEnabled(true);`. I'll continue testing with this update and inform of any updates.

RMR54:
I've updated and tested the updated order of `WPDXSetFusedTexBufferMax(1); ` call

PR #575, opened 2019-07-12
Add 64-bit support for USDigital encoders

dzenanz:
This is a work in progress, for which I need feedback.

During PlusLib generation step, I encounter `CMake Error: install(EXPORT "PlusLib" ...) includes target "vtkPlusDataCollection" which requires target "SEIDrv" that is not in the export set.` The solution files exist on disk, and can* be compiled and run. What is the best way to fix this? @adamrankin @lassoan @Sunderlandkyl 

*I encountered problems compiling:
```text
1>------ Build started: Project: SEIDrv, Configuration: Debug x64 ------
1>.NETFramework,Version=v4.0.AssemblyAttributes.cpp
1>C:\Temp\User\.NETFramework,Version=v4.0.AssemblyAttributes.cpp : fatal error C1083: Cannot open compiler generated file: 'Debug/.NETFramework,Version=v4.0.AssemblyAttributes.asm': No such file or directory
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(429,5): error MSB6006: "CL.exe" exited with code 1.
1>Done building project "SEIDrv.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 1 up-to-date, 0 skipped ==========
```
Some rebuilding steps solved the issue in 32-bit build (which I did in VS2017), but that did not help for 64-bit build (which I did in VS2019). Inspired by this [post](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/0bc20966-77e8-43cf-97d4-9c177daef17d/vs-2012-errors-compiling-dll-containing-files-with-the-same-name-fatal-error-c1083-cannot-open?forum=vcgeneral), changing `ASM List Location` from `Debug/` to `$(IntDir)` solved the problem.
Screenshot 2019-07-12 10 19 40
Does anyone know whether this is specific to CMake version? I am using 3.14.5 at the moment. @jamesobutler @RMR54 please report whether it build correctly for you, or whether you needed to resort to this workaround.


jamesobutler:
@dzenanz Yes, I'm also using CMake 3.14.5, Plus x64 build, VS2019 and running into the `CMake Error: install(EXPORT "PlusLib" ...) includes target "vtkPlusDataCollection" which requires target "SEIDrv" that is not in the export set.` that you mentioned.

jamesobutler:
I can also confirm that changing `ASM List Location` from `Debug/` to `$(IntDir)` did not work for my 64 bit build.

dzenanz:
> changing ASM List Location from Debug/ to $(IntDir) did not work

Did **not** work? Have you tried something else?

jamesobutler:
Ok, I've tried again. After changing to $IntDir and building again, it always reconfigures the SEIDrv project back to the previous setting.

dzenanz:
Aha, yes. Steps: configure+generate using CMake. Ignore the error message. Open Visual Studio. Change `ASM List Location` to `$(IntDir)`. Build. It should work.

But: next time CMake is re-run, `ASM List Location` will be reset to the default. If `SEIDrv` was not modified, the library will not need to be rebuilt, so the other stuff which depends on it will use the existing `SEIDrv.lib`.

jamesobutler:
Yeah I must be still doing something wrong. Yes, I do that change, build SEIDrv project and it then is successful. I then build the PlusBuild project again and it results in the previous failure again because it must've changed the `ASM list location` back to previous value.

dzenanz:
1. Build PlusBuild. It will fail due to CMake generate error in PlusLib. But PlusLib files will be there.
1. Open PlusLib.sln and build it. It will probably fail.
1. Change `ASM list location`. Do not run CMake any more!
1. Build PlusLib.sln again (not PlusBuild.sln!), which should be open now as you were just changing properties of SEIDrv project. It should build successfully.
1. Run PlusServer. For me, that entails setting working directory to `$(SolutionDir)../bin/$(Configuration)` and command arguments to `--config-file=C:\Dev\PlusGit\D32\PlusLibData\ConfigFiles\PlusDeviceSet_Server_MmfUSDigitalEncodersTracker_coreXY.xml`. Rotating the shaft changes the transform as observed in Slicer.

jamesobutler:
@dzenanz Ok, yes with following the above instructions I was able to build successfully. Thanks for the clarification! I was also able to test a mmf device with USDigital encoders and was able to see the transform change in Slicer when rotating the shaft.

dzenanz:
Thanks for feedback James!

During PlusLib generation step, I encounter `CMake Error: install(EXPORT "PlusLib" ...) includes target "vtkPlusDataCollection" which requires target "SEIDrv" that is not in the export set.` The solution files exist on disk, and can* be compiled and run. @adamrankin @lassoan @Sunderlandkyl What is the best way to fix this?

`SEIDrv.cpp` cannot be part of `vtkPlusDataCollection`, because it is not ordinary C++, but rather some Microsoft specific variant called `C++/CLI`. `vtkPlusDataCollection` does not compile if I set it to this language.

Sunderlandkyl:
Can it be compiled as an external project as part of PlusBuild?

dzenanz:
It could, but that seems like an overly complicated option to me. PlusLib already has multiple targets in it, can this be a new one, similarly to `vtkPlusUsSimulator` and `vtkPlusVolumeReconstruction`?

dzenanz:
This latest commit gets rid of the CMake warning. I still have to manually change ASM List Location, which is probably a CMake bug.

dzenanz:
CMake issue tracked [here](https://gitlab.kitware.com/cmake/cmake/issues/19480).

dzenanz:
I think this can be merged now. And it should work smoothly once https://gitlab.kitware.com/cmake/cmake/merge_requests/3569 makes it into a CMake release.

PR #579, opened 2019-07-25
DOC: Update VirtualCapture parameter default values

jamesobutler:
At this time, the default parameter values for Virtual Capture in the code do not match what the documentation says. This PR updates the documentation to match what is specified in the code.
https://github.com/PlusToolkit/PlusLib/blob/99a41bb4a34ee5a7b1d0d6307aee4442e19c7976/src/PlusDataCollection/VirtualDevices/vtkPlusVirtualCapture.cxx#L41

https://github.com/PlusToolkit/PlusLib/blob/99a41bb4a34ee5a7b1d0d6307aee4442e19c7976/src/PlusDataCollection/VirtualDevices/vtkPlusVirtualCapture.cxx#L60

### DIFF
--- a/src/Documentation/UserManual/DeviceVirtualCapture.dox
+++ b/src/Documentation/UserManual/DeviceVirtualCapture.dox
@@ -8,14 +8,14 @@ The file is saved as \ref FileSequenceMetafile format. If single file output for
 \section VirtualCaptureConfigSettings Device configuration settings
 
 - \xmlAtt \ref DeviceType "Type" = \c "VirtualCapture" \RequiredAtt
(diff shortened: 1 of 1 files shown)


jamesobutler:
@Sunderlandkyl How do we trigger an update to the documentation? This documentation update is in PlusLib but I guess http://perk-software.cs.queensu.ca/plus/doc/nightly/user/DeviceVirtualCapture.html only gets updated when there are PlusApp updates.

Sunderlandkyl:
Documentation should be rebuilt and updated nightly, but there's some issue with the automated tasks on the build machine. I'll run the documentation update manually and see if I can fix the issue.

Sunderlandkyl:
Documentation updated.

PR #596, opened 2019-09-08
Fix error in USDigital encoders during additional initialization

dzenanz:


dzenanz:
@jamesobutler please confirm that this resolves multi-initialization issue.

jamesobutler:
Yes, I've tested and 2 USDigitalEncodersTracker Plus devices can be initialized now.

PR #606, opened 2019-09-23
Do not throw exception if a COM port is already open

dzenanz:
This happens commonly if other devices are attached
via serial (COM) ports in addition to USDigital encoders.

### DIFF
--- a/src/SEIDrv/SEIDrv.cpp
+++ b/src/SEIDrv/SEIDrv.cpp
@@ -11,7 +11,7 @@ ref struct Globals
 
 void enumerateEncoders(long comPort, long mode)
 {
-    System::String^ comPortString = "COM"+ System::Convert::ToString(comPort);
+    System::String^ comPortString = "COM" + System::Convert::ToString(comPort);
 
     USDigital::SEIBusManager mgr;
     USDigital::SEIBus^ mSEIBus = mgr.GetBus(comPortString);
@@ -37,7 +37,14 @@ void enumerateEncodersAll(long mode)
         if (test != 0) // port exists
         {
             //std::cout << str << ": " << lpTargetPath << std::endl;
-            enumerateEncoders(i, mode);
+            try
+            {
+                enumerateEncoders(i, mode);
+            }
+            catch (System::UnauthorizedAccessException^)
+            {
+                // port is open by somebody else, ignore it
+            }
         }
     }
 }


dzenanz:
@jamesobutler please confirm that this fixes the issue with additional serial devices.

jamesobutler:
Yes, I'm able to successfully start a server that sets up GenericSerialDevices in addition to USDigitalEncodersTracker device.  This only works if the GenericSerialDevices have been specified above the USDigitalEncodersTracker device in the XML.  

Ideally the USDigitalEncodersTracker set up would not try to enumerate on a COM port that is being designated for a GenericSerialDevice. This code change is definitely some improvement as it works as long as the XML is configured in the correct way.

dzenanz:
A device cannot know what resources have been assigned to other devices. But we could introduce property `SerialPort="7"` which would tell it to look at a specified COM port.

jamesobutler:
US Digital Device Explorer which uses the new .NET framework can find and setup the connection to the correct COM port which has a US Digital device.  As seen below, COM11 and COM12 are devices that will be set up as GenericSerialDevices while COM7 is the SEI-USB device. This app can figure out that COM7 is the correct device. Is there nothing in the API that can do this?  Or is having the multi-initialization (https://github.com/PlusToolkit/PlusLib/commit/f94f3cc2d3c8536f1196be8f23f936a98b970916) support causing problems with the correct COM port initialization?
USDigitalDeviceExplorer

If the auto-discover behavior as in USDigitalDeviceExplorer is unable to be replicated when setting the device up in PLUS, then yes next best thing is to add the SerialPort property like is done for the GenericSerialDevices.

dzenanz:
As far as I can tell, our code does the same thing. I don't know why it disturbs the other serial devices.

lassoan:
Note that on Windows, com port names above 9 need to use the syntax `\\.\COM10`.

jamesobutler:
@dzenanz Since I couldn't find the document with all the API info, I did `ildasm USDSEI.dll` where I saw there is a `ScanDevices()` call. This sounds like it would do what USDigitalDeviceExplorer does.

dzenanz:
ScanDevices does something very similar to our code. It gets a list of available ports, then scans for devices on all of them. Disassembly:
```C#
public int ScanForDevices(string[] portsToScan, string[] deviceTypeFilterStrings, bool defaultBaudRateOnly)
{
  this.mPorts = portsToScan != null ? ((IEnumerable<string>) portsToScan).OrderBy<string, string>((Func<string, string>) (name => name)) : ((IEnumerable<string>) SerialPort.GetPortNames()).OrderBy<string, string>((Func<string, string>) (name => name));
  string str = deviceTypeFilterStrings == null ? string.Empty : string.Join(",", deviceTypeFilterStrings);
  this.Close();
  this.PostStatusUpdate("Scanning for connected devices using device filter = {" + str + "}", Priority.System);
  lock (this.mDeviceListLock)
    this.mDevices.Clear();
  foreach (string mPort in (IEnumerable<string>) this.mPorts)
  {
    try
    {
      SerialPort serialPort = new SerialPort(mPort);
      serialPort.Open();
      serialPort.Close();
      this.PostStatusUpdate("Scanning for device on " + mPort + "...", Priority.System);
      IDevice device1 = this.ScanForDeviceOnConnection(mPort, deviceTypeFilterStrings, defaultBaudRateOnly);
      if (device1 != null)
      {
        lock (this.mDeviceListLock)
        {
          this.mDevices.Add(device1);
          device1.OnClosed += new EventHandler(this.device_OnClosed);
          this.SetLatencyTimer(device1.Connection);
          if (this.OnConnectionChanged != null)
          {
            this.OnConnectionChanged(new DeviceManager.DeviceConnectionEventArgs(device1, device1.Connection, true));
            if (device1.GetType().GetInterface("IParentDevice") != (Type) null)
            {
              foreach (IDevice device2 in ((IParentDevice) device1).Devices)
              {
                this.mDevices.Add(device2);
                device2.OnClosed += new EventHandler(this.device_OnClosed);
                this.OnConnectionChanged(new DeviceManager.DeviceConnectionEventArgs(device2, device2.Connection, true));
              }
            }
          }
        }
      }
    }
    catch (Exception ex)
    {
      this.PostStatusUpdate("ScanForDevices Exception: Skip this port (" + mPort + ") :" + ex.Message, Priority.System);
    }
  }
  this.PostStatusUpdate("Scan complete.", Priority.System);
  return this.mDevices.Count;
}
```
The most likely problem is that once a COM port has been accessed, it cannot be accessed again immediately. See [this](https://stackoverflow.com/a/37839520/276168).

dzenanz:
I think this can be merged.

PR #632, opened 2019-11-22
BUG: Fix WinProbe not disconnecting all the way

jamesobutler:
The WinProbe device was not closing completely when disconnect was called. This uses 2 new calls as directed by the API developer. Tested with hardware and PlusServer will correctly close all the way now. 
 This requires UltraVisionAPI as of November 22nd 2019.

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -628,6 +628,8 @@ vtkPlusWinProbeVideoSource::~vtkPlusWinProbeVideoSource()
   {
     this->Disconnect();
   }
+  WPDispose();
+  WPDXDispose();
 }
 
 // ----------------------------------------------------------------------------


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -749,6 +749,8 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   {
     LOG_WARNING("Failed setting session pointer!");
     WPDisconnect();
+    WPDispose();
```
dzenanz: Maybe it is better to put these into `~vtkPlusWinProbeVideoSource` (class's destructor)?

jamesobutler: I tested just the new calls in the destructor as seen below, but it doesn't fix the disconnect problem.
```c++
vtkPlusWinProbeVideoSource::~vtkPlusWinProbeVideoSource()
{
  if(this->Connected)
  {
    this->Disconnect();
    WPDispose();
    WPDXDispose();
  }
}
```

dzenanz: `WPDispose();` and `WPDXDispose();` should not be within the `if` clause. If that doesn't help, then your original commit is good to go.

dzenanz: In the destructor, the device should already be disconnected.

jamesobutler: See latest change. Putting it outside of that `if` clause does work.

jamesobutler:
@adamrankin or @Sunderlandkyl, could you merge this PR? Thanks!

PR #664, opened 2020-02-20
ENH: Add initial ARFI support for WinProbe

adamaji:
This PR exposes some of the WinProbe ARFI API, and is largely work by @dzenanz . I've squashed ARFI related commits from https://github.com/dzenanz/PlusLib/commit/2d8f07e5be364e6ae1032ebbf645b7457d054995 to https://github.com/dzenanz/PlusLib/commit/a4e084ba6e17f74c53edab88ae4d531947ed0849 into the one ENH commit, and pulled out the commits from https://github.com/dzenanz/PlusLib/commit/f866d7ca3140c5d2c8cff940d61b83d0aff5b4ff until ARFI for the REFACTOR.

This is currently WIP as @tomekcz and I are confirming the returned data with a phantom. The implementation as it stands adds the ARFI data as one long 2d frame to the extra source which can be manipulated post acquisition, and can be changed later when the API is updated to allow changes to the number of 6x5 repetitions.

@dzenanz @jamesobutler , could you take a look?

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -222,7 +222,7 @@ int vtkPlusWinProbeVideoSource::MSecondsFromWidth(int32_t value)
 
 // ----------------------------------------------------------------------------
 vtkPlusWinProbeVideoSource* thisPtr = nullptr;
-int32_t quadBFCount = 0; // number of 4x beamformers
+int32_t quadBFCount = 1; // number of 4x beamformers
 
 //-----------------------------------------------------------------------------
 // This callback function is invoked after each frame is ready
@@ -333,15 +333,9 @@ void vtkPlusWinProbeVideoSource::FrameCallback(int length, char* data, char* hHe
   {
     int bSize = m_SSDecimation * m_PrimaryFrameSize[1] * m_PrimaryFrameSize[0];
     int timeblock = (4 / quadBFCount) * arfiGeometry->LineRepeatCount * sizeof(int32_t) * 30;
-    timeblock = timeblock / 2;  // for x8 bf engine? why isn't quadBFCount taking this into account?
     int arfiDataSize = arfiGeometry->SamplesPerLine * arfiGeometry->LineCount * arfiGeometry->LineRepeatCount * 30 * sizeof(int32_t);
-    if(length > bSize + arfiDataSize + timeblock){
-      LOG_WARNING("The size of the ARFI frame data is larger than expected.");
-    }
-    else if(length < bSize + arfiDataSize + timeblock){
-      LOG_WARNING("The size of the ARFI frame data is smaller than expected.");
-    }
-    frameSize[0] = length / sizeof(int32_t);  // we want to include the time data to be saved
+    assert(length == bSize + arfiDataSize + timeblock);
+    frameSize[0] = (bSize + arfiDataSize + timeblock) / sizeof(int32_t);  // we want to include all data to be saved
     frameSize[1] = 1;
     frameSize[2] = 1;
     if(frameSize != m_ExtraFrameSize)
@@ -826,7 +820,14 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   }
   WPSetSize(m_PrimaryFrameSize[0], m_PrimaryFrameSize[1]);
   SetMaxDmaTransferSize(0x100000);
-  quadBFCount = GetARFISSQuadBFCount();
+  if(::GetARFIIsX8BFEnabled())
+  {
+    quadBFCount = 2;
+  }
+  else
+  {
+    quadBFCount = 1;
+  }
   if(!m_UseDeviceFrameReconstruction)
   {
     WPDXSetIsGetSpatialCompoundedTexEnabled(true);
@@ -1230,7 +1231,7 @@ PlusStatus vtkPlusWinProbeVideoSource::SetExtraSourceMode(Mode mode)
     m_Mode = Mode::ARFI;
     unsigned bSize = m_SSDecimation * m_PrimaryFrameSize[1] * m_PrimaryFrameSize[0];
     unsigned arfiDataSize = 1024 * 16 * 64 * 30;
-    unsigned timeblockSize = (4 / 2) * 64 * 30;
+    unsigned timeblockSize = (4 / quadBFCount) * 64 * 30;
     m_ExtraFrameSize = { bSize + arfiDataSize + timeblockSize, 1, 1 };
     this->AdjustBufferSizes();
     std::vector<int32_t> zeroData(m_ExtraFrameSize[0] * m_ExtraFrameSize[1] * m_ExtraFrameSize[2], 0);
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -694,8 +694,8 @@ vtkPlusWinProbeVideoSource::~vtkPlusWinProbeVideoSource()
   {
     this->Disconnect();
   }
-  WPDispose();
   WPDXDispose();
+  WPDispose();
 }
 
 // ----------------------------------------------------------------------------
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -164,6 +164,8 @@ vtkPlusWinProbeVideoSource::Mode vtkPlusWinProbeVideoSource::StringToMode(std::s
   { return Mode::M; }
   else if(modeString == "PW")
   { return Mode::PW; }
+  else if(modeString == "ARFI")
+  { return Mode::ARFI; }
   else if(modeString == "CFD")
   { return Mode::CFD; }
   else
@@ -192,6 +194,9 @@ std::string vtkPlusWinProbeVideoSource::ModeToString(vtkPlusWinProbeVideoSource:
   case Mode::PW:
     return "PW";
     break;
+  case Mode::ARFI:
+    return "ARFI";
+    break;
   case Mode::CFD:
     return "CFD";
     break;
@@ -217,6 +222,7 @@ int vtkPlusWinProbeVideoSource::MSecondsFromWidth(int32_t value)
 
 // ----------------------------------------------------------------------------
 vtkPlusWinProbeVideoSource* thisPtr = nullptr;
+int32_t quadBFCount = 0; // number of 4x beamformers
 
 //-----------------------------------------------------------------------------
 // This callback function is invoked after each frame is ready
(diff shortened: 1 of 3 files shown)


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -711,6 +770,32 @@ PlusStatus vtkPlusWinProbeVideoSource::InternalConnect()
   {
     SetPWIsEnabled(true);
   }
+  if(m_Mode == Mode::ARFI)
+  {
+    SetARFIIsEnabled(true);
```
dzenanz: This whole block of code is a duplicate of, and should be replaced by `this->SetARFIEnabled(true);`

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -95,7 +95,7 @@ PlusStatus vtkPlusWinProbeVideoSource::ReadConfiguration(vtkXMLDataElement* root
       int mwidthSeconds = std::stoi(mwidthSeconds_string);
       if(mwidthSeconds > 0)
       {
-        m_MWidth = this->MWidthFromSeconds(mwidthSeconds);
+        m_ExtraFrameSize[1] = this->MWidthFromSeconds(mwidthSeconds); // [0] or [1]?
```
jamesobutler: ` // [0] or [1]?` as it related to this comment, I'm assuming this was figured out?

jamesobutler: @dzenanz Do you know about this?

dzenanz: Since this is the first change in the file, I might have put the comment there. I think I tested this, but testing again with M-mode would not hurt. The comment should be removed.

jamesobutler: I get a bunch of these errors when going from "M" mode to "BRF" mode which might be related to here? `|ERROR|115.700000| vtkPlusBuffer: Unable to add frame to video buffer - frame format doesn't match!`

jamesobutler: The warning directly before the error:
`|WARNING|028.599000| VideoDeviceLinearArray-ExtraSourceVideo: Frame format and buffer frame format does not match (expected frame size: 128x512x1  received: 1536x128x1)!| in C:\SP\Adam\SPR15_x64\PlusLib\src\PlusDataCollection\vtkPlusBuffer.cxx(198)`

adamaji: I've pushed a fix for this, there was a spot where the sizes passed to `adjustBufferSizes` were incorrect for brf mode.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -1125,37 +1199,90 @@ int32_t vtkPlusWinProbeVideoSource::GetSpatialCompoundCount()
...
+    LOG_DEBUG("GetARFIIsRFSampleDataCaptureEnabled: " << GetARFIIsRFSampleDataCaptureEnabled());
   }
+  SetPendingRecreateTables(true);
+  LOG_INFO("Mode changed to: " << ModeToString(mode));
+  m_Mode = mode;
+  return PLUS_SUCCESS;
 }
 
+//----------------------------------------------------------------------------
+void vtkPlusWinProbeVideoSource::SetBRFEnabled(bool value)
```
jamesobutler: I would suggest that this get changed to something like `BRFOn()`. The others would be `MModeOn()` and `ARFIOn()`.

adamaji: To preserve the setter/getter structure here and to allow the user to return to mode B after setting an extra source mode on, I made each of these call the new `SetExtraSourceMode` with `Mode::B` as an arg if `value` is false. I think this should be revisited if `SetMode` is looked at in the future

jamesobutler: It will have to be confirmed that you can be in `Mode::B` without a stream of errors due to no new frames coming across the extra source port.

adamaji: This is fine

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -1199,41 +1199,111 @@ int32_t vtkPlusWinProbeVideoSource::GetSpatialCompoundCount()
...
+    SetMAcousticLineIndex(m_MLineIndex);
+    ::SetMWidth(m_MWidth);
+    ::SetMAcousticLineCount(m_MAcousticLineCount);
+  }
+  else if(mode == Mode::ARFI)
+  {
+    m_Mode = Mode::ARFI;
+    unsigned bSize = m_SSDecimation * m_PrimaryFrameSize[1] * m_PrimaryFrameSize[0];
+    unsigned arfiDataSize = 1024 * 16 * 64 * 30;
+    unsigned timeblockSize = (4 / 2) * 64 * 30;
```
dzenanz: If we want to also use the 4x beamformer variants, this formula should incorporate `quadBFCount` variable and work for both 4x and 8x variants.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -478,12 +514,38 @@ void vtkPlusWinProbeVideoSource::FrameCallback(int length, char* data, char* hHe
     LOG_DEBUG("Frame ignored - B-mode source not defined. Got mode: " << std::hex << usMode);
     return;
   }
+  else if(usMode & ARFI)
+  {
+    for(unsigned i = 0; i < m_ExtraSources.size(); i++)
+    {
+      // send the ARFI data as one long 2D frame to be processed downstream
+      int32_t* tempData;
```
dzenanz: This declaration should be merged with assignment on line 526.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
adamaji: `quadBFCount` was returning 1 for me despite having an 8x engine. Does `SetARFISSQuadCount` need to be used for `GetARFISSQuadCount` to return the correct value? Would it be better to set this parameter from e.g., the DeviceConfig.xml?

dzenanz: I don't know. @malixg should know.

dzenanz: As this is hardware-dependent parameter, it is better not to put it in `DeviceConfig.xml`. We should query the hardware variant at run time.

jamesobutler: Use `GetARFIIsX8BFEnabled()` which will return a value of 1 for 8x board and a value of 0 for 4x board.

dzenanz:
I guess that using `GetARFIIsX8BFEnabled()` is the only thing left to do, and possibly clean up existing `quadBFCount` variable and associated code.

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -222,7 +222,7 @@ int vtkPlusWinProbeVideoSource::MSecondsFromWidth(int32_t value)
 
 // ----------------------------------------------------------------------------
 vtkPlusWinProbeVideoSource* thisPtr = nullptr;
-int32_t quadBFCount = 0; // number of 4x beamformers
+int32_t quadBFCount = 1; // number of 4x beamformers
```
jamesobutler: Comment could improve here as a value of 1 doesn't refer to 1 count of 4x beamformers based on how you use this variable.

dzenanz: The use of this variable looks correct to me.

adamaji:
>I guess that using GetARFIIsX8BFEnabled() is the only thing left to do, and possibly clean up existing quadBFCount variable and associated code.

Just pushed a commit for this

>In your ENH: Expose various WinProbe ARFI parameters commit, please also expose

@rat01 is looking at the remaining parameters currently, @tomekcz and i are looking at some returned data difference between the 8x and 4x engine

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
jamesobutler: The usage of the variable is correct. The comment is not as it makes it read as "1 number of 4x beamformers". Really it is like a True/False state for 8x/4x respectively so a change is needed to imrpove the specification that this specification of 1 refers to an 8x beamformer.

adamaji: the variable doesnt directly correspond to the ::getARFIIsx8Enabled() return, it's set to either 1 or 2 depending on that parameter, 2 for x8, so i think the comment makes sense here

jamesobutler: Ok sounds good

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -41,6 +41,10 @@ void vtkPlusWinProbeVideoSource::PrintSelf(ostream& os, vtkIndent indent)
   {
     os << indent << "FocalPointDepth" << i << ": " << m_FocalPointDepth[i] << std::endl;
   }
+  for(int i = 0; i < 6; i++)
+  {
+    os << indent << "ARFIFocalPointDepth" << i << ": " << m_ARFIFocalPointDepth[i] << std::endl;
```
dzenanz: Since you are printing it to console/log, maybe you should read/write it to config file?

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -102,6 +106,7 @@ PlusStatus vtkPlusWinProbeVideoSource::ReadConfiguration(vtkXMLDataElement* root
 
   deviceConfig->GetVectorAttribute("TimeGainCompensation", 8, m_TimeGainCompensation);
   deviceConfig->GetVectorAttribute("FocalPointDepth", 4, m_FocalPointDepth);
+  deviceConfig->GetVectorAttribute("ARFIFocalPointDepth", 6, m_ARFIFocalPointDepth);
```
dzenanz: Yes, this is what I was referring to in previous review's comment.

PR #670, opened 2020-02-28
pass new function parameters to clariusInitListener

zoez526:
According to https://github.com/clariusdev/listener/blob/master/src/include/listen.h , the previous version of vtkPlusClarius needed to add button callback function and Framewidth, FrameHeight as arguments to the clariusInitListener() function.


src/PlusDataCollection/Clarius/vtkPlusClarius.cxx
```diff
@@ -828,6 +846,7 @@ PlusStatus vtkPlusClarius::WritePosesToCsv(const ClariusImageInfo* nfo, int npos
       posInfo += (std::to_string(systemTime) + ",");
       posInfo += (std::to_string(convertedTime) + ",");
       posInfo += (std::to_string(nfo->tm) + ",");
+      posInfo += (std::to_string(nfo->micronsPerPixel) + ",");
```
zoez526:
```suggestion

```

PR #672, opened 2020-03-03
ENH: Update WinProbe ARFI focal point depth method

jamesobutler:
With recent changes to the UltraVision (WinProbe) API, these changes to Set/Get ARFI Focal point depth are required.

This was tested with hardware to confirm the change.


jamesobutler:
@Sunderlandkyl @adamrankin This is ready. Could you merge this for me?

PR #673, opened 2020-03-03
ENH: Expose WinProbe ARFIPushOffset

jamesobutler:
This exposes ARFIPushOffset for the WinProbe device so that PLUS can set it to different values other than the default.

This was tested with hardware to confirm the change.


jamesobutler:
@Sunderlandkyl @adamrankin This is ready. Could you merge this one for me too?

PR #699, opened 2020-06-26
Otitrack valid tracking status

hgueziri:
- Fixed CMakeLists paths for natnet for 32-bit compilation.

- Added a check for valid tracking using the first bit of rigidbody.params.
Before, to check if a tracked tool is valid (in view), we test MeanError != 0. This fails to tell if a tool is out of view. When setting SendValidTransformsOnly="FALSE" in the config file, out of view tools have the latest valid transform with ToolStatus Ok.


src/PlusDataCollection/OptiTrack/vtkPlusOptiTrack.cxx
```diff
@@ -378,7 +378,10 @@ void vtkPlusOptiTrack::vtkInternal::InternalCallback(sFrameOfMocapData* data, vo
     rigidBodyToTrackerMatrix->Identity();
     sRigidBodyData currentRigidBody = rigidBodies[rigidBodyId];
 
-    if (currentRigidBody.MeanError != 0)
+    // Check if tool is in view
+    bool bTrackingValid = currentRigidBody.params & 0x01; 
+
+    if ((currentRigidBody.MeanError != 0) & (bTrackingValid))
```
adamrankin, line 384: This changes the behaviour to only send valid transforms. Plus supports the ability to send invalid transforms with the status flag set. If you want to pass this information to the client, you'd need to modify line 406.

adamrankin:
If you'd like to continue, I'd ask you to make the changes to vtkPlusOptiTrack.cxx to not change the current behaviour, but to set  the status flag on send. That way it's up to the client program to either ignore or use an invalid transform.

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -246,9 +246,11 @@ IF(PLUS_USE_OPTITRACK)
   # Collect all the MotiveAPI libraries
   SET(PLATFORM_NAME_SUFFIX "")
   SET(LIBRARIES_PATH_SUFFIX "lib32/")
+  SET(OPTITRACK_NATNET_PATH_SUFFIX "x86/")
   IF (CMAKE_HOST_WIN32 AND CMAKE_CL_64 )
     SET(PLATFORM_NAME_SUFFIX "x64")
     SET(LIBRARIES_PATH_SUFFIX "")
+	SET(OPTITRACK_NATNET_PATH_SUFFIX "x64/")
```
adamrankin, line 253: Please convert to spaces instead of a tab.

src/PlusDataCollection/OptiTrack/vtkPlusOptiTrack.cxx
hgueziri: I'm not sure. Invalid transforms are sent through line 412. This is the whole point of this PR. 
I think ```currentRigidBody.MeanError != 0``` is supposed to detect if a marker is visible (?), but is always true, even when markers are not visible (tested with Optitrack Duo camera and SendValidTransformsOnly="FALSE"). I left it in case there is another purpose of checking meanError != 0?

hgueziri:
@adamrankin thanks for reviewing the code. I've made the changes.
Though, when the tool is not visible the transform matrix is being computed unnecessarily (from the last valid frame). Initially, I was trying to avoid that by putting the condition earlier. What do you think?

adamrankin:
> Though, when the tool is not visible the transform matrix is being computed unnecessarily (from the last valid frame). Initially, I was trying to avoid that by putting the condition earlier. What do you think?

Let's leave it computing the transform. There may be other reasons to compute an invalid transform

Issue #719, opened 2020-09-22
Do not use TerminateThread to stop threads


lassoan:
This change can introduce hard-to-detect regressions in PlusServer: https://github.com/PlusToolkit/PlusLib/pull/718

TerminateThread is a nuclear bomb - it is essentially a forced partial crash of the application. We should only use it as a very last resort, if ever, as it can put the application into an inconsistent state. It might be OK to do forced termination if there does not seem to be a chance for the thread to recover (for example, if a thread does not respond for 10 seconds; it may be still better to kill that thread instead of making the whole application hang indefinitely).

However, in this case, most likely the thread exits as requested, but for some reason vtkMultiThreader still sees the stopped thread as active.

lassoan:
@adamaji Could you debug into this and see vtkMultiThreader did not see the thread as stopped before? Did the thread did not exit successfully? Or maybe we should just call CloseHandle(threadID) so that vtkMultiThreader can properly detect that the thread is completed?

lassoan:
I've just checked and vtkMultiThreader::TerminateThread does not seem to actually terminate the thread but sets a flag that requests it to stop and waits for it to complete. So, the thread is actually not terminated forcefully (no risk of making the application to get into inconsistent state).

@adamaji can you check if keeping the original implementation (without https://github.com/PlusToolkit/PlusLib/pull/718) and just calling CloseHandle for the complete thread fixes the issue that you had?

adamaji:
>can you check if keeping the original implementation (without #718) and just calling CloseHandle for the complete thread fixes the issue that you had?

Yes, I'll look into this

adamaji:
From earlier debugging, the problem was that vtkMultiThreader did not see the thread as stopped, and thus does not free up its slots for `SpawnedThreadActiveFlag/Lock`. The DataReceiverThread in OpenIGTLinkServer does actually finish, setting `ClientData->DataReceiverActive.second` to false.

Using `CloseHandle` here doesn't work because we don't have any reference to the actual process id of the thread, the `DataReceiverThreadId` is just an index in the vtkMultiThreader. That's why I used vtkMultiThreader::TerminateThread in the first place -- it doesn't seem like there are other methods of freeing up the spawned thread variables in the multi threader without deconstructing entirely. We also cannot get the process id from the multi threader as those are out of scope / protected.

lassoan:
Thanks for checking this. The followings should be sufficient then:

1. Revert back
[...]

PR #720, opened 2020-09-22
BUG: Clarify usage of vtkMultiThreader::TerminateThread in OpenIGTLinkServer

adamaji:
This commit addresses requested changes (#719) for a prior commit f1fab02 which cleans up used thread IDs in the multithreader for OpenIGTLinkServer. The fixed behavior is maintained.

cc @lassoan

### DIFF
--- a/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
+++ b/src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
@@ -1128,7 +1128,7 @@ void vtkPlusOpenIGTLinkServer::DisconnectClient(int clientId)
         {
           continue;
         }
-        if (this->Threader->IsThreadActive(clientIterator->DataReceiverThreadId))
+        if (clientIterator->DataReceiverThreadId >= 0)
         {
           if (clientIterator->DataReceiverActive.second)
           {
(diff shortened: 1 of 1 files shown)


src/PlusServer/vtkPlusOpenIGTLinkServer.cxx
```diff
@@ -1137,8 +1137,12 @@ void vtkPlusOpenIGTLinkServer::DisconnectClient(int clientId)
           }
           else
           {
-            // stop thread
+            // thread stopped
+            // We still need to call vtkMultiThreader::TerminateThread, not to terminate the thread (as it is already stopped)
```
adamaji, line 1141: I suppose saying the thread is already stopped here is half-true, as the logic in vtkPlusOpenIGTLinkServer::DataReceiverThread has finished, but at this point we haven't called CloseHandle/Join on the thread. vtkMultiThreader::TerminateThread ends up calling CloseHandle/Join. I'm not sure if it's necessary to close the handle explicitly once the thread logic has completed / to mention this in the comment?

lassoan (commented):
The thread is actually already stopped by then (thread function finished, ensured by checking the active flag), so the comment is accurate.

lassoan (approved):
Thank you, this looks good. The clarification was needed because Plus and vtkMultiThreader has slightly different, somewhat redundant mechanism for requesting stopping of a thread and waiting for a thread to complete.

PR #727, opened 2020-10-09
COMP: Enable compilation for VTK >= 8.9

adamrankin:
Not yet tested with 8.2


adamrankin:
8.2 compiles and runs fine

Does not package with vtk 9

PR #731, opened 2020-10-26
ENH: Allow external access for adding custom headers from VirtualCapture

adamaji:
I'd like to add this change so that custom header fields can be set from a virtual capture object at the time the file is being written. I wasn't sure how else to do this without making the associated tracked frame list public, so if there's an easier/better solution please let me know. Thanks!


adamrankin:
Could you give an example of how you want to use this?

adamaji:
Sure. For our use case we have a custom clone of PlusServer which handles starting/stopping recording and saving of files across multiple ultrasound devices. We'd like for the saved files that are written via VirtualCapture devices in our clone to include custom header information about e.g., scan-specific subject information, and for that information to be written at acquisition time instead of needing to reopen the file later to modify the header. The proposed method here allows us to set a header field after starting recording for one of our VirtualCapture devices.

dzenanz (approved):

PR #736, opened 2020-11-13
ENH: Add initial support for Andor high sensitivity CCD cameras

adamaji:
This PR adds base support for interfacing with Andor CCD cameras. Most of the changes here made by @dzenanz 

The corresponding PlusBuild changes currently live at https://github.com/dzenanz/PlusBuild and we'll make a PR for that too.

CC @dzenanz @jamesobutler @jrojasUNC can you give these changes another look? Squashing here seemed to make the most sense as one "initial commit"


src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+/*!
+ \class vtkPlusAndorCamera
+ \brief Class for acquiring images from Andor cameras
+
+ Requires PLUS_USE_ANDOR_CAMERA option in CMake.
+ Requires the Andor SDK (SDK provided by Andor).
+
+ \ingroup PlusLibDataCollection.
+*/
+class vtkPlusDataCollectionExport vtkPlusAndorCamera: public vtkPlusDevice
```
adamrankin, line 22: The naming convention for video devices is vtk<DeviceName>VideoSource. In this case it should be vtkPlusAndorVideoSource in so named .h/.cxx files

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+   * 0 Fully Auto
+   * 1 Permanently Open
+   * 2 Permanently Closed
+   * 4 Open for FVB series
+   * 5 Open for any series
+   *
+   * For an external shutter: Output TTL high signal to open shutter.
+   */
+  PlusStatus SetShutter(int shutter);
+  int GetShutter();
```
adamrankin, line 58: This is just a preference, but should this be an enum?

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+
+  /*! Acquisition mode. Valid values:
+   * 1 Single Scan
+   * 2 Accumulate
+   * 3 Kinetics
+   * 4 Fast Kinetics
+   * 5 Run till abort
+   */
+  PlusStatus SetAcquisitionMode(int acquisitionMode);
+  int GetAcquisitionMode();
```
adamrankin, line 84: Same, enum?

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+
+  /*! Readout mode. Valid values:
+   * 0 Full Vertical Binning
+   * 1 Multi-Track
+   * 2 Random-Track
+   * 3 Single-Track
+   * 4 Image
+   */
+  PlusStatus SetReadMode(int setReadMode);
+  int GetReadMode();
```
adamrankin, line 94: Same, enum?

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+   * 0. Internal
+   * 1. External
+   * 6. External Start
+   * 7. External Exposure (Bulb)
+   * 9. External FVB EM (only valid for EM Newton models in FVB mode)
+   * 10. Software Trigger
+   * 12. External Charge Shifting
+   */
+  PlusStatus SetTriggerMode(int triggerMode);
+  int GetTriggerMode();
```
adamrankin, line 106: Same, enum?

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+
+  /*! Lowest temperature at which it is safe to shut down the camera. */
+  PlusStatus SetSafeTemperature(int safeTemp);
+  int GetSafeTemperature();
+
+  /*! Get the current temperature of the camera in degrees celsius. */
+  float GetCurrentTemperature();
+
+  /*! Paths to additive and multiplicative bias correction images. */
+  PlusStatus SetBiasCorrectionImage(std::string biasFilePath);
```
adamrankin, line 120: Even though the compiler would probably optimize it, should this be const &?

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
```diff
@@ -0,0 +1,266 @@
...
+  /*! Device-specific recording start */
+  PlusStatus InternalStartRecording() override;
+
+  /*! Device-specific recording stop */
+  PlusStatus InternalStopRecording() override;
+
+  /*! Initialize vtkPlusAndorCamera */
+  PlusStatus InitializeAndorCamera();
+
+  using DataSourceArray = std::vector<vtkPlusDataSource*>;
```
adamrankin, line 180: TIL a new using syntax. @lassoan @Sunderlandkyl Should this be moved to the top of the class declaration (similar to typedef's)?

src/PlusDataCollection/CMakeLists.txt
```diff
@@ -2336,6 +2337,46 @@ IF(PLUS_USE_V4L2)
     )
 ENDIF()
 
+# --------------------------------------------------------------------------
```
adamrankin, line 2340: Could you also create a folder grouping (lower down in this file?)

src/PlusDataCollection/Testing/vtkAndorCameraTest.cxx
```diff
@@ -0,0 +1,223 @@
...
+  See License.txt for details.
+=========================================================Plus=header=end*/
+
+/*!
+  \file vtkAndorCameraTest.cxx
+  \brief Test basic connection to the Andor camera
+
+  If the --rendering-off switch is defined then the connection is established, images are
+  transferred for a few seconds, then the connection is closed (useful for automatic testing).
+  If the --rendering-off switch is not defined then the live ultrasound image is displayed
```
jrojasUNC, line 13: not ultrasound but optical.

src/PlusDataCollection/Andor/vtkPlusAndorCamera.h
dzenanz: It should be `const`, but perhaps not `const &`.

src/PlusDataCollection/Andor/vtkPlusAndorCamera.cxx
```diff
@@ -0,0 +1,956 @@
...
+// ----------------------------------------------------------------------------
+unsigned int vtkPlusAndorCamera::checkStatus(unsigned int returnStatus, std::string functionName)
+{
+  if(returnStatus == DRV_SUCCESS)
+  {
+    return returnStatus;
+  }
+  else if(returnStatus == DRV_NOT_INITIALIZED)
+  {
+    LOG_ERROR("Failed AndorSDK operation: " << functionName            \
```
dzenanz, line 786: Is line continuation (`\`) necessary? Even if so, the spaces are not.

src/PlusDataCollection/Andor/vtkPlusAndorCamera.cxx
```diff
@@ -0,0 +1,956 @@
...
+
+  cv::subtract(floatImage, cvBiasCorrection, floatImage, cv::noArray(), CV_32FC1);
+  LOG_INFO("Applied constant bias correction");
+
+  // Divide the image by the 32-bit floating point correction image
+  cv::divide(floatImage, cvFlatCorrection, floatImage, 1, CV_32FC1);
+  LOG_INFO("Applied multiplicative flat correction");
+
+  // OpenCV's lens distortion correction
+  cv::undistort(floatImage, result, cvCameraIntrinsics, cvDistanceCoefficients);
```
jrojasUNC, line 416: Based on recent conversations, the lens distortion correction should be applied before flat correction.

src/PlusDataCollection/Andor/vtkPlusAndorCamera.cxx
```diff
@@ -0,0 +1,956 @@
...
+  {
+    ApplyFrameCorrections();
+    AddFrameToDataSource(GrayRectified);
+  }
+
+  return PLUS_SUCCESS;
+}
+
+// ----------------------------------------------------------------------------
+PlusStatus vtkPlusAndorCamera::AcquireBiasFrame(std::string biasFilePath, int binning, int vsSpeed, int hsSpeed)
```
jrojasUNC, line 456: We are going to do a single calibration image which includes both the read bias and dark current. However, this might change in the future so I think we should have the flexibility to take both type of images (Bias- shutter closed, exposure of 0 sec, dark charge- shutter closed, variable exposure) so exposure should be an input to this function. Furthermore, it might be helpful to acquire an image with the shutter open for troubleshooting so the shutter mode should also be an input.

jrojasUNC: It might be appropriate to call this function AcquireCorrectionFrame since it could collect both bias and dark current images.

src/PlusDataCollection/Andor/vtkPlusAndorCamera.cxx
```diff
@@ -0,0 +1,956 @@
...
+{
+  LOG_TRACE("vtkPlusAndorCamera::InternalConnect");
+  if(this->InitializeAndorCamera() != PLUS_SUCCESS)
+  {
+    LOG_ERROR("Andor camera failed to initialize.");
+    return PLUS_FAIL;
+  }
+
+  this->GetVideoSourcesByPortName("BLIraw", BLIraw);
+  this->GetVideoSourcesByPortName("BLIrectified", BLIrectified);
```
jrojasUNC, line 224: Should the corrected ports be called BLICorrected and GrayCorrected or BLIProcessed and GrayProcessed? I think "rectified" implies an operation that is not happening.

jrojasUNC (changes requested):

adamaji:
@adamrankin thanks for the review! I've made some changes as squashable commits based on your comments, as well as from @dzenanz and @jrojasUNC .

Re: https://github.com/PlusToolkit/PlusLib/pull/736#discussion_r522540003, I've opted to leave that where it is since its usage is pretty much like how `typedef` can be used in that context, and there are other instances in the repo where `typedef` is used like that and not moved to the top of the class declaration. 

I've also opened https://github.com/PlusToolkit/PlusLibData/pull/36 and https://github.com/PlusToolkit/PlusBuild/pull/69 as companion PRs.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -57,9 +57,9 @@ PlusStatus vtkPlusAndorVideoSource::ReadConfiguration(vtkXMLDataElement* rootCon
 
   Shutter = static_cast<ShutterMode>(std::atoi(deviceConfig->GetAttribute("Shutter")));
   m_AcquisitionMode = static_cast<vtkPlusAndorVideoSource::AcquisitionMode>(std::atoi(deviceConfig->GetAttribute("AcquisitionMode")));
+  m_ReadMode = static_cast<vtkPlusAndorVideoSource::ReadMode>(std::atoi(deviceConfig->GetAttribute("ReadMode")));
```
dzenanz, line 60: If "ReadMode" is absent from configuration file, this will default to 0 meaning FullVerticalBinning. This is undesirable.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -468,6 +468,10 @@ PlusStatus vtkPlusAndorVideoSource::SetBiasCorrectionImage(const std::string bia
   try
   {
     cvBiasCorrection = cv::imread(biasFilePath, cv::IMREAD_GRAYSCALE);
```
dzenanz: cv::imread throws exceptions, so this might not be necessary. Have you run into a case where this is needed?

adamaji: Yes, running with the bias/flat paths set to a path that can't be found will silently fail the imread, and the resulting cv::Mat will be empty. This is just for easier debugging in that case

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
adamaji: would something like this be better?
```
if(deviceConfig->GetAttribute("ReadMode")
{
  m_ReadMode = static_cast<vtkPlusAndorVideoSource::ReadMode>(std::atoi(deviceConfig->GetAttribute("ReadMode")));
}
else
{
  m_ReadMode = ReadMode::Image;
}
```

and same with `m_AcquisitionMode` and `m_TriggerMode`?

dzenanz: Yes. It is more verbose, but more convenient to use. Maybe even like:
```
const char * readModeString = deviceConfig->GetAttribute("ReadMode");
if(readModeString)
{
  m_ReadMode = static_cast<vtkPlusAndorVideoSource::ReadMode>(std::atoi(readModeString));
}
else
{
  m_ReadMode = ReadMode::Image;
}
```

dzenanz (approved):

PR #745, opened 2020-12-09
Andor camera startup and close improvements

jamesobutler:
This PR is a series of commits to improve understanding of the Andor camera startup and close connection. `Initialize` is now only called if there are available cameras.  This provides a better understanding of what goes wrong when `Initialize` fails. If fails to initializes it will return early instead of failing to do the various SDK commands. Additional logging of the Camera model and Serial Number on startup also provides better understanding of what was connected.

Previously if the Camera cooler was set to return ambient, it would wait to warm up to the Safe Temperature before calling `ShutDown`. This is unnecessary for most of their most common cameras and appears to be mostly for their older cameras. I have removed this requirement for now. If someone has an old camera and use this class they can implement the necessary change.  I clarified this fact with an Andor employee prior to proceeding with this change. Their response is quoted below and I've included this in the changed files as well.

> Only Classic, ICCD and cameras with a fibre attached must have their cooling and warming up controlled at a particular rate.
> For everything else you can just call ShutDown and the camera will safely return to room temperature.
> Classic systems are cameras that use our original PCI controller cards eg CCI-010 or CCI-001.

All of these changes was tested with an Andor CCD camera.

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
@@ -233,6 +233,20 @@ PlusStatus vtkPlusAndorVideoSource::InitializeAndorCamera()
     return PLUS_FAIL;
   }
 
+  char headModel[_MAX_PATH];
+  unsigned headModelResult = checkStatus(GetHeadModel(headModel), "GetHeadModel");
+  if(headModelResult == DRV_SUCCESS)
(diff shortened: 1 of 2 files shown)


adamrankin:
Does this need any confirmation/testing before I merge?

adamaji (approved):
No, this should be good to go.

dzenanz (commented):
Looks good.

PR #750, opened 2020-12-15
ENH: Add bad pixel correction

jrojasUNC:
CCD cameras usually have "bad pixels" these are either dead pixels, pixels with significantly lower intensities than neighbors, or pixels whose intensity does not scale with exposure time. Resolution cells (for example 4x4 pixels if bin=4) with more than 20% bad pixels are replaced by the median of surrounding cells. 

I have included the DeviceConfig.xml that I used and all of the supporting files here [](url)
[Bad Pixel Correction supporting files.zip](https://github.com/PlusToolkit/PlusLib/files/5692714/Bad.Pixel.Correction.supporting.files.zip)

I randomly made up an bad pixel image for testing and it has bad pixels here:
image

If the algorithm works the image may look basically the same as without the algorithm because the camera being used may not have many dead pixels so I suggest first changing line 642 of vtkPlusAndorVideoSource to `cvIMG.at<ushort>(resolutionCellIndexY, resolutionCellIndexX) = 0;`. This will write an image where the resolution cells that have bad pixels will be set to zero:
image

These images show that the algorithm is correctly identifying the location of the bad pixels from the bad pixel image. It also shows that the algorithm will not correct (in this case make 0) resolution cells where less than 25% of the pixels are bad (bin 4 and bin 8 images).

Reverting the change to make the cells with bad pixels 0 (original state) will produce images that look like ones above but without any black pixels and the regions where there are bad pixels look the same are their surroundings.

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
@@ -18,6 +18,7 @@ vtkStandardNewMacro(vtkPlusAndorVideoSource);
 // put these here so there is no public dependence on OpenCV
 cv::Mat cvCameraIntrinsics;
 cv::Mat cvDistanceCoefficients;
+cv::Mat cvBadPixelImage;
 cv::Mat cvFlatCorrection;
 cv::Mat cvBiasDarkCorrection;
 
@@ -46,6 +47,7 @@ void vtkPlusAndorVideoSource::PrintSelf(ostream& os, vtkIndent indent)
   os << indent << "UseFrameCorrections: " << UseFrameCorrections << std::endl;
   os << indent << "FlatCorrection: " << flatCorrection << std::endl;
   os << indent << "BiasDarkCorrection: " << biasDarkCorrection << std::endl;
+  os << indent << "BadPixelCorrection: " << badPixelCorrection << std::endl;
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jamesobutler:
cc: @dzenanz for review

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.h
```diff
@@ -141,6 +143,11 @@ class vtkPlusDataCollectionExport vtkPlusAndorVideoSource: public vtkPlusDevice
   float GetCurrentTemperature();
 
   /*! Paths to additive and multiplicative bias+dark charge correction images. */
```
dzenanz, line 145: This docstring should be updated

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -539,11 +544,114 @@ void vtkPlusAndorVideoSource::AddFrameToDataSource(DataSourceArray& ds)
...
+    uint resolutionCellIndex, badIndices;
+    uint resolutionCellIndexX, resolutionCellIndexY, minX, maxX, minY, maxY, badPixelCount;
+    for (int i = 0; i < badIndicesXY.size(); i++)
+    {
+      badPixelCount = 0;
+      resolutionCellIndexX = badIndicesXY[i].x / binning;
+      resolutionCellIndexY = badIndicesXY[i].y / binning;
+      resolutionCellIndex = frameSize[1] * resolutionCellIndexY + resolutionCellIndexX;
+
+      if (std::find(resolutionCellsWithBadPixels.begin(), resolutionCellsWithBadPixels.end(), resolutionCellIndex) == resolutionCellsWithBadPixels.end())
```
dzenanz, line 563: I believe this logic could be simplified a little.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -539,11 +544,114 @@ void vtkPlusAndorVideoSource::AddFrameToDataSource(DataSourceArray& ds)
...
+// ----------------------------------------------------------------------------
+void vtkPlusAndorVideoSource::CorrectBadPixels(int binning, cv::Mat& cvIMG)
+{
+    std::vector<cv::Point> badIndicesXY;
+    cv::findNonZero(cvBadPixelImage, badIndicesXY);
+
+    std::vector<uint> resolutionCellsWithBadPixels, resolutionCellsToCorrect;
+    uint resolutionCellIndex, badIndices;
+    uint resolutionCellIndexX, resolutionCellIndexY, minX, maxX, minY, maxY, badPixelCount;
+    for (int i = 0; i < badIndicesXY.size(); i++)
```
dzenanz, line 556: This should be done in `SetBadPixelCorrectionImage`, not once per frame.

dzenanz (changes requested):
Juan, I will make some changes to this and push them to my fork. Then you can integrate into your fork or we could make a new PR to PlusLib.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
dzenanz: Now I realize this depends on binning too.

dzenanz:
@jrojasUNC please check modifications in [my fork](https://github.com/dzenanz/PlusLib/tree/add-bad-pixel-correction).

PR #751, opened 2020-12-17
ENH: Add bad pixel correction

dzenanz:
CCD cameras usually have "bad pixels" these are either dead pixels, pixels with significantly lower intensities than neighbors, or pixels whose intensity does not scale with exposure time. Resolution cells (for example 4x4 pixels if bin=4) with more than 20% bad pixels are replaced by the median of surrounding cells.

Supersedes and closes #750.


jrojasUNC (approved):
Tested it and it looks good.

PR #752, opened 2020-12-18
ENH: add cosmic ray correction

dzenanz:


adamrankin:
Just let us know when to merge.

dzenanz:
I think this is ready to merge, but I defer to @jrojasUNC.

dzenanz:
... and Juan is on vacation until January. I guess this will keep pending for two more weeks.

jrojasUNC (approved):
Sorry for the delay. I just tested this and it looks good.

PR #756, opened 2021-01-05
Expose setters for undistortion parameters in Andor device

adamaji:
This PR exposes setters/getters for camera intrinsics and distance coefficients for the Andor device, since these parameters must be changed for different binning settings.

Other small changes are converting the units for processed Andor images into counts/seconds and fixing a bug where correction images are loaded as the wrong type.

cc @jrojasUNC @dzenanz for review

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
@@ -670,7 +670,7 @@ void vtkPlusAndorVideoSource::ApplyCosmicRayCorrection(int bin, cv::Mat& floatIm
 }
 
 // ----------------------------------------------------------------------------
-void vtkPlusAndorVideoSource::ApplyFrameCorrections(int binning)
+void vtkPlusAndorVideoSource::ApplyFrameCorrections(int binning, float exposureTime)
 {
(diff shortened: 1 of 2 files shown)


src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -993,6 +993,32 @@ bool vtkPlusAndorVideoSource::GetUseCosmicRayCorrection()
   return this->UseCosmicRayCorrection;
 }
 
+// ----------------------------------------------------------------------------
+PlusStatus vtkPlusAndorVideoSource::SetCameraIntrinsics(double* intrinsics)
```
dzenanz, line 1000: Style-wise and from type safety standpoint, it might be better to use `std::vector<double>` or `std::array<double, 9>`. Possibly also update ivar to be the same type.

dzenanz (approved):
Looks good, in-line suggestion.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
adamaji: Thanks, pushed a change for this. I left the internal var as is, because I wasn't sure how to handle `Set/GetVectorAttribute` nicely in `Read/WriteConfiguration` which was complaining about expecting `double *`s.

dzenanz (approved):

PR #765, opened 2021-01-18
Andor index value updates

jamesobutler:
This exposes a few more methods from the Andor SDK to get the actual HSSpeed and VSSpeed values.  There has been some getters in this Andor PlusLib class, but they were returning the index values. Andor's SDK uses index values to "Set" the value, but then the corresponding Andor SDK Getter returns the actual value.  These changes should make it clearer about if the index or the actual value is being used.

To Maintainers: It would be great if you can "Rebase and Merge" this PR to include the multiple commits here. This will help me manage corresponding updates in a private repo.

cc: @adamaji, @dzenanz

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.h
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.h
@@ -78,6 +78,7 @@ public:
 
   /*! Set index to use in the vertical shift speed table. */
   PlusStatus SetVSSpeedIndex(int index);
+  int GetVSSpeedIndex();
 
   /*! Get the actual VS Speed in microseconds per pixel shift for the current VSSpeed index. */
   float GetVSSpeed();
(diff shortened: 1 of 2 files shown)


src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.h
```diff
@@ -73,6 +73,9 @@ class vtkPlusDataCollectionExport vtkPlusAndorVideoSource: public vtkPlusDevice
   /*! Set horizontal shift speed. */
   PlusStatus SetHSSpeed(int type, int index);
```
dzenanz, line 74: No renaming this to *Index?

jamesobutler: The framework that relies on the PlusConfig xml for setting values on startup (HSSpeed="0 1" will call SetHSSpeed(0, 1)) was the reason I didn't change the name here because then the config file would say have to say `HSSpeedIndex="0 1"` and only one of those values in the array is the actual index. The first value is actually the type of output amplification.

dzenanz (approved):

jamesobutler:
Thanks for the review! @adamrankin or @Sunderlandkyl can you merge. As mentioned in the original message:

> To Maintainers: It would be great if you can "Rebase and Merge" this PR to include the multiple commits here. This will help me manage corresponding updates in a private repo.

PR #766, opened 2021-01-18
ENH: Allow Andor device to set frame fields with acquisition parameters

adamaji:
This PR makes the Andor device set some frame fields with acquisition parameters (exposure time, binning) on each frame acquisition. Also exposes an `ImageToReference` transform setter so that a tracker doesn't have to be used for setting the frame field transform.


adamaji:
cc @dzenanz for review

dzenanz (approved):

PR #775, opened 2021-02-05
ENH: Several changes to improve usability of Andor device

adamaji:
There are several changes here which make interfacing with the Andor devices in downstream code easier, such as added checks to see whether the acquisition thread is running, as well as some small bug fixes.

cc @dzenanz and @jamesobutler for review,
and @jrojasUNC for commits related to the bad pixel processing and the counts/sec -> counts change


jamesobutler (approved):
Tested with hardware and this looks good. Thanks!

The revert to not enforce the bad pixel correction image to current the image size makes sense as indeed it should be the main detector size for how it is used. I should've caught this when reviewing Dzenan's last PR recently.

PR #776, opened 2021-02-05
Capistrano improvements

dzenanz:


dzenanz:
@jamesobutler @adamaji I think this PR can be reviewed/merged. These are bug fixes and minor improvements.

src/PlusServer/Commands/vtkPlusGetUsParameterCommand.cxx
```diff
@@ -300,7 +300,7 @@ vtkPlusUsDevice* vtkPlusGetUsParameterCommand::GetUsDevice()
     // No ultrasound device id is specified, auto-detect the first one and use that
     for (DeviceCollectionConstIterator it = dataCollector->GetDeviceConstIteratorBegin(); it != dataCollector->GetDeviceConstIteratorEnd(); ++it)
     {
-      usDevice = vtkPlusUsDevice::SafeDownCast(*it);
+      usDevice = dynamic_cast<vtkPlusUsDevice*>(*it);
```
dzenanz, line 303: @Sunderlandkyl and @adamrankin it would be good if you double-checked this change. The old line fails to cast `vtkPlusCapistranoVideoSource` into its parent class, `vtkPlusUsDevice`.

Sunderlandkyl: I think that the old line failed because this line: https://github.com/PlusToolkit/PlusLib/blob/master/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.h#L28

Should be:
```vtkTypeMacro(vtkPlusCapistranoVideoSource, vtkPlusUsDevice);```

Comment (dzenanz) [reply src/PlusServer/Commands/vtkPlusGetUsParameterCommand.cxx:303]: You are right! I will amend the PR.

File: src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.h [hunk 2]
```diff
@@ -26,22 +26,22 @@ class vtkPlusDataCollectionExport vtkPlusCapistranoVideoSource: public vtkPlusUs
   /*! Constructor for a smart pointer of this class*/
   static vtkPlusCapistranoVideoSource* New();
   vtkTypeMacro(vtkPlusCapistranoVideoSource, vtkPlusDevice);
```
adamrankin, line 28: Would it be possible to change the base type to be vtkPlusUsDevice

dzenanz: I just did that, and it resolves the issue.

PR #784, opened 2021-02-09
BUG: Andor processed frames not normalized from 32 to 16 bit correctly

adamaji:
Processed Andor frames were not being appropriately scaled to [0, 2^16) when converting from CV_32FC1 to CV_16UC1.

cc @jrojasUNC @dzenanz for review

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
@@ -9,6 +9,7 @@ See License.txt for details.
 #include "vtkPlusAndorVideoSource.h"
 #include "ATMCD32D.h"
 #include "igtlOSUtil.h" // for Sleep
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
I am fine with the code. I expect that you looked at the intensity values computed and that you liked them :D

adamaji:
Thanks Adam! @dzenanz yes haha, for history here's an example of the images I was looking at (of a glowing watch):

Before this PR ( values range from [0, 4] ):
image

With these changes ( values range from [0, 2^16) ):
image

dzenanz:
If two images have different peak intensities, do they both end up with 2^16-1 after this normalization? If yes, is that what you want?

adamaji:
Hmm, yes that's right, and since we're measuring counts that's probably incorrect. So perhaps NORM_RELATIVE is the correct operation here? @jrojasUNC ?

PR #785, opened 2021-02-16
Set single flat image but resize for different binnings for Andor acquisitions

jrojasUNC:
This changes the code so that instead of setting a different flat image for the different binnings, a single flat image is set and resized to match the size of the binned image. The resized image is saved in a container so the resizing doesn't have to happen every time an image is acquired.

the intensity at the corners of the corrected image is increased due to the correction image used:
image

@adamaji @dzenanz


adamaji (approved):
Looked over a couple images of a light phantom in various configurations (covered by paper, in the corner of the fov) with @jrojasUNC to verify functionality. There are some differences in count ranges due to other applied corrections, but now the ranges are appropriately scaled to the respective captures.

(ex. unprocessed and processed images)
image
image

dzenanz (approved):

PR #787, opened 2021-02-17
BUG: Andor drops frames after filtered timestamp stops being valid

adamaji:
This PR fixes an issue with the Andor device where after a set number of captures (around 20 for one data source), no new frames would be added to the data source. This is because the computed filtered timestamp was found as not "`filteredTimestampProbablyValid`". The major change here is to just send the clock as our timestamp. This device generally acquires single frames a few seconds apart, so this should be fine.

A question for maintainers though, what's the motivation for returning `PLUS_SUCCESS` when the created filtered timestamp is found to be probably invalid, and thus the frame isn't added to the data source?
https://github.com/PlusToolkit/PlusLib/blob/master/src/PlusDataCollection/vtkPlusBuffer.cxx#L341-L345

### DIFF
--- a/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
+++ b/src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
@@ -580,7 +580,7 @@ void vtkPlusAndorVideoSource::AddFrameToDataSource(DataSourceArray& ds)
                       1, US_IMG_BRIGHTNESS, 0,
                       this->FrameNumber,
                       currentTime,
-                      UNDEFINED_TIMESTAMP,
+                      currentTime,  // just use the unfiltered timestamp so we don't drop frames
                       &this->CustomFields
                      ) != PLUS_SUCCESS)
     {


dzenanz:
Filtering for a device can be disabled by setting it to 1 or 0, I think.

src/PlusDataCollection/Andor/vtkPlusAndorVideoSource.cxx
```diff
@@ -591,6 +591,11 @@ void vtkPlusAndorVideoSource::AddFrameToDataSource(DataSourceArray& ds)
       LOG_INFO("Success adding item to AndorCamera video source " << ds[i]->GetSourceId());
     }
   }
+  unsigned status = checkStatus(FreeInternalMemory(), "FreeInternalMemory");
```
dzenanz, line 594: Why do this? It will need to be allocated for the next frame, probably.

adamaji: It was one of the things we tried when first debugging the issue and was just left in, but I can take it out. It shouldn't be related to the main issue at hand.

>Filtering for a device can be disabled by setting it to 1 or 0, I think.

Which setting are you thinking of? Is it a member of vtkPlusDevice?

dzenanz: AveragedItemsForFiltering, in e.g. http://perk-software.cs.queensu.ca/plus/doc/nightly/dev/DeviceCapistrano.html

adamaji: Hmm I see so when this is set <2 the unfiltered timestamp is just used as the "filtered" timestamp instead. Do you think this should just be a change in our device config? I think that would mean that anytime that `AveragedItemsForFiltering` is set to anything other than 0 or 1 in the config, then acquisition will still fail (if somebody else uses the device for example), so shouldn't the forced unfiltered timestamp be done here as well?

dzenanz: Somebody else could be using higher frame rates? But since the device is meant for longer exposures, doing it with this patch makes more sense.

dzenanz (approved):

PR #789, opened 2021-02-18
BUG: Andor shutter transfer time backwards success check

adamaji:
Small oversight in checking for success for the Andor device shutter transfer time setters.

cc @dzenanz @jamesobutler


jamesobutler (approved):
Yes, this looks good

PR #805, opened 2021-05-19
Fix some Wobbler parameters not set correctly on startup from config

dzenanz:


jamesobutler (approved):
I tested with hardware and was able to see a difference in the image when changing JitterCompensation value in the config file with these changes. Previously I was seeing no difference.

jamesobutler:
@Sunderlandkyl or @adamrankin can you provide final review/merge? Corresponding PR is also at https://github.com/PlusToolkit/PlusLibData/pull/41 that also needs final review/merge. Thanks!

src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
```diff
@@ -100,7 +100,7 @@ class vtkPlusCapistranoVideoSource::vtkInternal
     int   b                   = tgcMin;
     float m                   = (float)(tgcMax - tgcMin) / samplesPerLine;
 
-    for (int x = 0; x < samplesPerLine; x++)
+    for(int x = 0; x < samplesPerLine; x++)
```
Sunderlandkyl, line 103: Is the removal of the space between for/if and the bracked caused by the AStyle from this page?: https://github.com/PlusToolkit/PlusBuild/blob/master/Docs/BuildInstructionsWindows.md?

If it is, then we should update it so that the spaces are not removed.

dzenanz: Yes, I installed AStyle only for Plus, and followed the instructions. But I did it long ago, maybe the style was updated since?

Sunderlandkyl: Doesn't look like it's been updated recently.
If you could fix the spacing in this PR, then I'll integrate it and update the AStyle on both https://github.com/PlusToolkit/PlusBuild/blob/master/Docs/BuildInstructionsWindows.md and https://github.com/PlusToolkit/plustoolkit.github.io/blob/c420073809345ab94de5c3a2333f8b56fe2bb486/developersguide.md.

dzenanz: Done. Please let me know when AStyle config is updated so I can copy-paste it into my Visual Studio settings.

Sunderlandkyl: Thanks! I'll let you once it's updated.

Sunderlandkyl: I've created .astylerc with the updated config in the PlusLib repository: https://github.com/PlusToolkit/PlusLib/blob/master/.astylerc

PR #812, opened 2021-05-27
Resolving last issues with remote control of Capistrano device

dzenanz:
This should make it possible to work without warnings even if not all parameters are specified in the config file. There will be companion PlusLibData PR.

CC: @jamesobutler.

### DIFF
--- a/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
+++ b/src/PlusDataCollection/Capistrano/vtkPlusCapistranoVideoSource.cxx
@@ -677,7 +677,6 @@ void vtkPlusCapistranoVideoSource::PrintSelf(ostream& os, vtkIndent indent)
   os << indent << "ClockDivider: " << ClockDivider << std::endl;
   os << indent << "CineBuffers: " << CineBuffers << std::endl;
   os << indent << "SampleFrequency: " << SampleFrequency << std::endl;
-  os << indent << "PulseFrequency: " << PulseFrequency << std::endl;
   os << indent << "WobbleRate: " << (int)GetWobbleRate() << std::endl;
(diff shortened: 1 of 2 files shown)


adamrankin:
Does this need to be tested with hardware?

dzenanz:
I already tested it, and it works for me. But I would prefer if @jamesobutler or @adamaji confirmed it works for them too.

jamesobutler:
Yes I should test with Hardware. I will be able to do this next week after the Federal holiday on Monday.

jamesobutler (approved):
I was able to test using the UltrasoundRemoteControl module in 3D Slicer. I was able to observe Capistrano imaging depth and frequency appropriately updating. 

For UltrasoundRemoteControl I had to make a minor update to allow higher frequencies as it currently supports 2-5 MHz as defined [here](https://github.com/openigtlink/SlicerOpenIGTLink/blob/7f02269c73ebcfccd2edf508cd01440d016e1ab9/UltrasoundRemoteControl/UltrasoundRemoteControl.py#L75-L78) while the Capistrano API supports up to 50 MHz as an option. It is also difficult to use the sliders for that module, but that's an area of improvement for that module and not PlusLib. @Sunderlandkyl Have you used this module recently and identified possible areas of improvement?

Sunderlandkyl:
I haven't used it recently, but there are several projects that use qSlicerUltrasoundDoubleParameterSlider to change ultrasound parameters. What difficulty did you have with the Sliders?

jamesobutler:
Frequency slider was limited to 2-5 while my device supports 10-50. It also supports discrete values rather than continuous for frequency and imaging depth. Typing the value into the spinboxes next the to the slider would often get rejected due to typing a 2 digit number resulting in it get rejected when the first digit was entered. I really had to drag the slider and then let it go over a valid value to trigger the update.

PR #837, opened 2021-07-22
BUG: Incorrect spacing frame field from winprobe rf capture

adamaji:
Minor bug in the winprobe device which saved incorrect element spacings to rf extra source captures. Also cleaned up what seem to be some copy paste typos in conditionals.

cc @jamesobutler for review


jamesobutler (approved):
I’ve confirmed the fix with hardware.

@Sunderlandkyl or @adamrankin can you integrate? Thanks!

PR #840, opened 2021-08-09
ENH: Clarify ARFIPushConfigurationString usage

jamesobutler:
This includes some clarifications when using a X4BF WinProbe device when doing ARFI. The X4BF continues to use the older API such as ARFIPushOffset, while the X8BF can use the custom specification of that ARFIPushConfigurationString supports.

Dual support will be maintained until X4BF devices are nearly phased out. X4BF devices have not been in production for awhile , but are still used occasionally.

cc: @dzenanz

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -1762,6 +1762,34 @@ int32_t vtkPlusWinProbeVideoSource::GetARFIStopSample()
 }
 
 //----------------------------------------------------------------------------
+void vtkPlusWinProbeVideoSource::SetARFIPushOffset(int32_t value)
+{
+  if(Connected)
+  {
+    if(quadBFCount != 1)
+    {
+      LOG_WARNING("ARFI Push offset is only used by X4BF devices. Use SetARFIPushConfigurationString instead.");
+    }
+    ::SetARFIPushOffset(value);
+    SetPendingRecreateTables(true);
+  }
+}
+
+//----------------------------------------------------------------------------
+int32_t vtkPlusWinProbeVideoSource::GetARFIPushOffset()
+{
+  if(Connected)
+  {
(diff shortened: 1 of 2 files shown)


src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
```diff
@@ -1762,20 +1762,31 @@ int32_t vtkPlusWinProbeVideoSource::GetARFIStopSample()
...
-      return PLUS_FAIL;
+      if(m_ARFIPushConfigurationCount == 30)
+      {
+        LOG_WARNING("A X4BF will use an API hardcoded 30 push configuration rather than the custom specified configuration that was set.")
+      }
+      else
+      {
+        LOG_ERROR("A X4BF requires an ARFI Push Configuration string to have 30 pushes so that the buffers are sized correctly. "
+                  "An appropriate configuration length will be set instead.");
+        pushConfiguration = "1,40,48;1,48,56;1,56,64;1,64,72;1,72,80;1,80,88;"
```
dzenanz, line 1808: This variable is assigned to, but not used. As it is a parameter that is passed by value, this will not be seen from outside this method.

dzenanz (approved):

src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
jamesobutler: As in it should be the following at time of set? Currently `m_ARFIPushConfigurationString` is only updated at time of `GetARFIPushConfigurationString`.
```C++
m_ARFIPushConfigurationString = pushConfiguration
WPSetARFIPushConfigurationString(pushConfiguration.c_str());
```

dzenanz: I didn't see it being used on line 1787. Sorry for the noise.

jamesobutler:
@adamrankin could you merge this PR? Thanks!

PR #866, opened 2021-12-02
ENH: Update WinProbe for latest ARFI API

jamesobutler:
This PR updates the WinProbe device based on the latest updates to the device ARFI API.

The API has dropped ARFI support for X4BF engines. ARFI requires X8BF. ARFI LineRepeatCount has been replaced with PrePushLineRepeatCount and PostPushLineRepeatCount.

@adamaji Can you provide review here with details of testing of these changes with hardware?


adamaji (approved):
Things look good here. I've run these changes with hardware without issue -- I compared ARFI images of known-stiffness phantoms acquired with this build and the master branch (using the default pre/post push line repeat counts to match the hardcoded line repeat count on the master branch) and found them to be similar enough to pass.

jamesobutler:
@adamrankin or @Sunderlandkyl can you proceed with merging? Thanks!

PR #886, opened 2022-01-07
ENH: Improve US digital connection for coreXY setups

jamesobutler:
Makes sure to fail if either of the 2 required encoders for the coreXY configuration are not present.

cc: @adamaji who can confirm my testing with hardware


adamaji (approved):

jamesobutler:
@adamrankin or @Sunderlandkyl Can you help with merging of this PR? Thanks!

PR #941, opened 2022-05-02
ENH: Expose overall gain adjustment for WinProbe device

adamaji:
Small change for exposing overall time gain compensation adjustment for WinProbe devices.

cc @jamesobutler or @nathanbmnt for review


nathanbmnt (approved):
lgtm

PR #947, opened 2022-05-12
ENH Added ability to set Atracsys frame options from config file

atracsys-sbt:
Frame options allow to control the output from Atracsys devices. For optimization purposes, one may want to control the maximum number of events, 2d/3d fiducials and markers that can be provided per frame in the output container.
So far, these options were hardcoded in the ftkSetFrameOptions call. For example, the max number of markers was set to 4 while Atracsys trackers are able to track more.


adamrankin:
does this need any hardware testing or review?

atracsys-sbt:
It has been successfully tested on a ftk500 and a stk180.

PR #948, opened 2022-05-13
ENH: Separate frame polling and callback methods for WP device

adamaji:
Originally the Winprobe device class would use `WPDXGetFusedTex` to grab post-processed B-Mode frames, but triggered from the device's new frame callback which pulls from a separate pool of pre-processed frame data. The intention of this API call is to use it in a polling manner. These changes separate that call into `InternalUpdate` such that the device now can appropriately poll for post-processed frames while still using the callback for other data (RF), where the polled data does not depend on the callback (except for setting an initial timestamp offset).

cc @jamesobutler (and maybe @dzenanz if you are available?) for review


dzenanz (approved):
I didn't see problems with this. Did you try it, and does it work?

PR #984, opened 2022-09-20
Fixes #983 use singleshot to give Qt event subsystem time to correctly establish signals/slots and connections

### LINKED ISSUE #983: DeviceSetSelectorWidget doesn't emit DeviceSetSelected signal at launch
Hi,
I am developping an application that includes a DeviceSetSelectorWidget and I would like to display some kind of preview of the content of the configuration file upon selection. Connecting a slot function to QPlusDeviceSetSelectorWidget::DeviceSetSelected does the trick nicely. However, upon relaunching the application, the last selection is stored and automatically set in the configuration list. In that case, the signal is not emitted.
I think it would be helpful to have the signal emitted at launch or at least to have access in read-only to the currently config file path from outside the class (so far only the description is accessible via QPlusDeviceSetSelectorWidget::GetSelectedDeviceSetDescription).
Thanks

adamrankin:


adamrankin:
@Sunderlandkyl Can you take a look?

adamrankin:
Sorry about the autoformatting changes, the only code change is the singleshot line and #includes

Sunderlandkyl:
Looks good to me.

Some of the packages still use Qt 5.7 (StealthLink/Ultrasonix), which doesn't accept std::chrono, so I changed it to an int to fix the compile error for those builds.

PR #997, opened 2022-10-19
ENH: Stop Winprobe from scanning if xducer disconnected from engine

adamaji:
This change is to stop the winprobe engine from pulsing if the transducer is suddenly disconnected.

cc: @jamesobutler for review


jamesobutler (approved):
Yes I have confirmed with hardware that this change works. Thanks.

cc: @adamrankin or @Sunderlandkyl to help with merging this change.

Issue #1004, opened 2022-10-28
Dependencies not aligned with 3D Slicer


jamesobutler:
3D Slicer 5.0.0 uses ITK 5.3rc03 and soon to be release Slicer 5.2.0 uses ITK 5.3rc04. PlusBuild is still specifying ITK 5.2.0 which results in the following warnings about potential incompatibility when connecting a PlusServer to Slicer (see output below). Generally PlusToolkit has tried to stay aligned with 3D Slicer regarding external toolkits such as ITK as previously discussed in https://github.com/PlusToolkit/PlusBuild/pull/52.

```
WARNING: In C:\SP\SP22R\itk\Modules\Core\Common\src\itkObjectFactoryBase.cxx, line 597
Possible incompatible factory load:
Running itk version :
itk version 5.2.0
Loaded factory version:
itk version 5.3.0
Loading factory:
C:/Program Files/SonoEQ 2.0.0-2022-10-14/bin/../lib/SonoEQ-5.1/ITKFactories\MRMLIDIOPlugin.dll
```

ITK 5.3 requires C++14 at minimum and an updated CMake minimum version.
- https://github.com/Slicer/Slicer/commit/bc2b3f4049ef32486cf43ad16903820fc8767f27
- https://github.com/Slicer/Slicer/commit/664fecaf8d70acdfaba30a473fe35931a4999e47

3D Slicer further moved to C++17 in https://github.com/Slicer/Slicer/commit/dedb6c04d5f987d168e58dfda067717a7039b439. I see that some devices in Plus such as ClariusOEM require C++17 as seen [here](https://github.com/PlusToolkit/PlusBuild/blob/master/CMakeLists.txt#L16-L23).

Due to this C++ minimum standard upgrade I have also observed issues building OpenCV where PlusBuild [specifies version 3.4.7](https://github.com/PlusToolkit/PlusBuild/blob/master/SuperBuild/External_OpenCV.cmake) which might be too old for a new C++ standard. The SlicerOpenCV extension builds and uses OpenCV [4.5.5](https://github.com/Slicer/SlicerOpenCV/blob/3f0bbd87c2e821020944d805405fe603874477c1/SuperBuild/External_OpenCV.cmake#L35).

[...]

PR #1011, opened 2022-11-21
Removing old support in preparation of dependency updates

jamesobutler:
re https://github.com/PlusToolkit/PlusLib/issues/1004 to align standards and toolkits used in latest 3D Slicer

This is a PR associated with https://github.com/PlusToolkit/PlusBuild/pull/82

cc: @adamrankin, @Sunderlandkyl, @lassoan


lassoan:
It would simplify things if we could remove these from the source code, but we still need Ultrasonix and StealthLink support. I think keeping the latest Plus version compatible is still the easiest solution. If we display meaningful error messages at configuration time when Ultrasonix and StealthLink is enabled and some incompatible options are chosen then we could avoid unexpected build errors.

jamesobutler:
Is Ultrasonix or Stealthlink supported on a version of Windows currently supported?

lassoan:
Yes, the client (that is in Plus) work on latest Windows 11 for both. The OS in the device is WinXP embedded for Ultrasonix (that must be connected to the network via firewall and direct internet access is disabled), StealthStation is using some recent Linux version.

jamesobutler:
## Stealthlink
> StealthStation is using some recent Linux version.

Ok so StealthStation isn’t stuck using Windows 7/8? That’s what the below text at https://plustoolkit.github.io/download made it seem like.

> PlusApp-…-StealthLink-Win32: Package to be used for connecting to Medtronic navigation systems through StealthLink interface on 32/64-bit, Windows 7 and Windows 8 systems. Requires StealthLink license. Package is available on request.

## Ultrasonix
> The OS in the device is WinXP embedded for Ultrasonix (that must be connected to the network via firewall and direct internet access is disabled),

So one of the dependency build configuration options must still support Windows XP? The below text is from https://plustoolkit.github.io/download

> Supported on Windows 7 or Windows XP Embedded:
> 
> PlusApp-…-Ultrasonix-5.7-Win32: Package to be used on Ultrasonix systems (running on Windows XP embedded or Windows 7 operating system); running Ultrasonix Exam software version between 5.7.0 and 6.0.2. MicronTracker and Microsoft Media Foundation imaging devices are not supported. Executables are 32-bit applications.

lassoan:
I've updated the documentation and download page to make things more clear (we only support Windows 10 and 11; except the Ultrasonix package, which we support on XP and XPe). Probably in a couple of years we can wind down Ultrasonix support, but it is still the only ultrasound system that has built-in electromagnetic tracking sensor in the transducer and has a research interface.

PR #1013, opened 2022-11-22
DOC: Update example PlusVersion output

jamesobutler:
This updates the example output of PlusVersion as seen at http://perk-software.cs.queensu.ca/plus/doc/nightly/user/ApplicationPlusVersion.html to a newer example output. Now the example is a little bit closer to what a user is likely to see when using the latest version of Plus.

Example output was generated from the PlusApp-2.9.0.20221116-Win64.exe package downloaded from https://plustoolkit.github.io/download


Sunderlandkyl:
Thanks!

PR #1014, opened 2022-11-23
ENH: Atracsys wrapper now also streams all the marker data and events

atracsys-sbt:
This is a significant update for the Atracsys wrapper. The Marker and Fiducial classes have been rewritten to store all the data provided by the Atracsys API. This data, in addition to events, are now streamed in a custom field along side each tool source.
The max number of fiducials and markers per frame have also both been bumped to 256 to handle many fiducials at the same time.


Sunderlandkyl:
Does this allow the same Plus build to compile with support for both spryTrack and fusionTrack?

atracsys-sbt:
> Does this allow the same Plus build to compile with support for both spryTrack and fusionTrack?

Unfortunately not at the moment. The spryTrack and fusionTrack use different SDKs. I'll look into the possibility of compiling with both simultaneously though. I agree that would be helpful not to have two versions.

PR #1047, opened 2023-02-06
ENH: Add command to read frame rate from channel data source

adamaji:
This PR adds a vtkPlusCommand for retrieving the frame rate from a channel's data source.

The command input structure looks like
```
<Command Name="GetFrameRate" ChannelId="MyChannelId" />
```
The output structure looks like
```
<CommandReply FrameRate=123.456 FrameRateStdDev=0.123456 />
```

This is helpful for our group since we work with devices with varying frame rates based on changes to certain imaging parameters, and we'd like to monitor those frame rate changes through Slicer.

cc @jamesobutler who's reviewed these changes with hardware


jamesobutler (approved):
Yes I have tested with hardware and confirms that I can return the frame rate of the source. 

cc: @Sunderlandkyl or @adamrankin for help with integrating this PR. Thanks!

PR #1052, opened 2023-02-15
ENH: Update to latest US Digital library version

jamesobutler:
This updates the USDigital SEI library version from an old 2013/2014 version to the latest 2023 version. This is maintained at https://github.com/PlusToolkit/PlusLib/tree/master/src/SEIDrv. The only change to note is the addition of a `devicesExpected` parameter when the library is searching for devices on an SEI bus. This was introduced by USDigital and now the vtkPlusUSDigitalEncodersTracker.cxx class uses it appropriately.

USDigital.dll 1.0.31.0 (2014) -> 1.1.0.6 (2023)
USDSEI.dll 1.0.28.0 (2013) -> 1.0.31.8 (2023)

These updated DLLs were downloaded from the official location at https://www.usdigital.com/support/resources/downloads/software/sei-software/

I have tested this with the US Digital hardware and had success.
cc: @nhjohnston for confirmation of hardware testing


nhjohnston (approved):
Confirmed that these PlusLib changes work; tested with USDigital Hardware

PR #1054, opened 2023-02-17
SEI disconnection and retry connections

jamesobutler:
This adds some changes to the SEIDrv to explicitly close the SEI bus at time of close. It also adds retry connection attempts based on a USDigital example to address certain issues of failed connections where the connection started soon after another connection ended.

cc: @adamaji


adamaji:
Can confirm I've tested this with hardware without issue and have observed the retry connection process succeeding - cc @Sunderlandkyl or @adamrankin can you help with merging this PR? Thanks!

PR #1085, opened 2023-05-04
ENH: Expose ARFI InterSet and InterPush delays

jamesobutler:
This exposes a couple new parameters (InterSet and InterPush delays) for the WinProbe device.

cc: @hannah-morilak for an approving review based on hardware testing

`BUG: Fix Winprobe ARFI cmds failing in debug builds` is for compatibility with the latest WinProbe API which has the InterSet and InterPush parameters.

### DIFF
--- a/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
+++ b/src/PlusDataCollection/WinProbe/vtkPlusWinProbeVideoSource.cxx
@@ -1902,6 +1902,11 @@ PlusStatus vtkPlusWinProbeVideoSource::SetARFILineTimer(uint16_t propertyValue)
 
   if(Connected)
   {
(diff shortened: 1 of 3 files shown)


hannah-morilak (approved):
I have tested with hardware and can confirm that `GetARFIInterSetDelay`, `SetARFIInterSetDelay`, `GetARFIInterPushDelay` and `SetARFIInterPushDelay` all work. 
Additionally, SetARFIInterSetDelay is successfully erroring when attempting to set above 62500, SetARFIInterPushDelay is erroring when attempting to set above 250 and SetARFILineTimer is erroring for above 2000

jamesobutler:
Thanks @hannah-morilak!

cc: @Sunderlandkyl Can you help with the integration of this PR? Thanks!

PR #1103, opened 2023-08-02
ULL IACTEC SYNCHRO - Combined video 

fmarcanoull:
ULL IACTEC SYNCHRO - Combined video (DAQ System + MMF camera)


fmarcanoull:
Good morning. Here I'm sending a code of a new composite device (DAQMMFCombinedVideo) to be merged. This code allows capturing thermal and visible frames simultaneously for MMF and DAQ setup.   Thanks.

cpinter:
A first comment I have is that please squash the commits (making sure there is no merge commit either), and prefix it with `ENH:`.
I think it would facilitate review if you explained the reason behind such a combined device.

@lassoan @Sunderlandkyl I noticed that there are several PRs in PlusLib, but no reaction. There is activity in PlusLib so I'd assume it is probably an oversight or lack of time. If it is possible please review, I haven't worked on Plus in the past 8 years or so, so my utility here is questionable. Thank you!

Sunderlandkyl:
Yes, sorry it was an oversight on my part.

I have the same comments as Csaba. what is the use case for this device that can't be accomplished by creating a config file that includes both the separate DAQ and MMF devices?

fmarcanoull:
The reason was trying to reduce the difference between videostreams coming from each device, implementing the same approach used for stereo cameras (e.g. OvrvisionPro), and so, having a setup of cameras synchronized at frame level, to show the same video scene in different spectra with minimum and fixed time delay. Making this at config file level, has shown in our tests there was a noticeable delay between the two videostreams. Some of the questions I asked to Csaba in the past were in that direction, to know if perfect video synchronization was guaranteed just implementing the composite device using only the config file.  In the end, the current implementation has provided to me the best synchronization between the cameras. Surely, I'm not aware of some other alternative that could be implemented without making an ad hoc device (with all its limitations) that includes both cameras, and in that case, I'd appreciate a lot if that information could be provided to me.

adamrankin:
Closely synced video feeds is an issue.

The downside to this approach is the explosive potential of dual-video devices to be maintained.

fmarcanoull:
Closed.

fmarcanoull:
The point is that, as mentioned in the thread, this development is not suitable for incorporation into the general library, so I am keeping it as a development for private use by the requester of this simultaneous acquisition feature for both sensors. Due to this, I have decided to close the request that was still open.

PR #1115, opened 2023-09-19
ENH: Implement support for firmware revision querying and BX2 command support

adamrankin:
If the device has firmware G.003 or higher, we can support some of the newer commands including BX2

Eventually we can implement STREAM support, which separates data flow and control flow to 2 ports


adamrankin:
See https://github.com/PlusToolkit/ndicapi/pull/37

adamrankin:
Data averaging feature is not working as intended. Aborting for now

PR #1126, opened 2023-10-31
ENH: Implement support for firmware revision querying and BX2 command support

adamrankin:
Intelligently detect the available firmware version on the device and utilize new commands if available
Enable support for hardware data averaging performed on NDI device, update timeout accordingly (see API documentation)


Sunderlandkyl (approved):
LGTM!

PR #1127, opened 2023-11-03
COMP: Fix compilation on Linux with Atracsys

errollgarner:
I don't know if this line is required on Windows. I'm assuming it is (I can't test currently), so I moved it in the windows section. If not, then it could be removed entirely. On Linux it adds an non-existent object to all dependent build.make files which causes compilation to fail.


Sunderlandkyl:
Are you using the latest PlusBuild with the AtracsysSDK_HOME environment variable defined?

https://github.com/PlusToolkit/PlusBuild/blob/9d9c1c20a753731deede9f3517f628a863ac83c2/Modules/FindAtracsysSDK.cmake#L41

errollgarner:
I confirm that using the latest master branch of PlusBuild on Linux Mint (Ubuntu 22.04) results in:
```
make[5]: *** No rule to make target '/home/user/PLTools/Atracsys/linux/ftk/lib/fusionTrack64', needed by '/home/user/build_plusbuild/bin/libvtkPlusDataCollection.so'.  Stop.
make[4]: *** [CMakeFiles/Makefile2:1967: src/PlusDataCollection/CMakeFiles/vtkPlusDataCollection.dir/all] Error 2
make[3]: *** [Makefile:146: all] Error 2
make[2]: *** [CMakeFiles/PlusLib.dir/build.make:92: PlusLib-prefix/src/PlusLib-stamp/PlusLib-build] Error 2
make[1]: *** [CMakeFiles/Makefile2:1013: CMakeFiles/PlusLib.dir/all] Error 2
make: *** [Makefile:101: all] Error 2
```
And removing the offending line of the CMakeLists enables compilation.

errollgarner:
> Are you using the latest PlusBuild with the AtracsysSDK_HOME environment variable defined?
> 
> https://github.com/PlusToolkit/PlusBuild/blob/9d9c1c20a753731deede9f3517f628a863ac83c2/Modules/FindAtracsysSDK.cmake#L41

The AtracsysSDK_HOME variable was indeed not set at the environment level though. Why is this a new requirement? What does it need to be set to?

Sunderlandkyl:
It was introduced as part of this pull request to update for the latest SDK changes.
I assume that it needs to point to the install directory of the SDK. @atracsys-sbt can clarify.

atracsys-sbt:
Indeed, since the latest version of our SDK, we now rely on environment variables for our SDK installation. In Windows, this is automatically done in the Registry. In Linux, unfortunately, this variable needs to be manually set, as described in our user manual.

PR #1129, opened 2023-11-09
ENH: Speed up PlusVersion check that includes WinProbe device

jamesobutler:
This moves some initialization and destruction of resources to the internal connect/disconnect calls instead of the device constructor and destructor. This is based on recommendations per https://plustoolkit.github.io/devicecode

"Initialization and destruction of resources is performed during YourDevice::InternalConnect and YourDevice::InternalDisconnect respectively."

cc: @matt-harmody-pki


matt-harmody-pki (commented):
I've tested these changes on hardware and observed no issues with connection or data acquisition.

PR #1150, opened 2024-01-10
BUG: Fix Capistrano device failing to connect after power cycle with cSDK2023 drivers

jamesobutler:
This fixes an issue where the Capistrano device would fail to connect the first time after it was power cycled. This will now keep checking if the imaging engine board is present as it waits for the device to re-enumerate after firmware is uploaded to the board during initial connection after being power cycle.


adamrankin:
Do you want any hardware testing before merging?

jamesobutler:
I was to confirm the fix through hardware testing and @nhjohnston was able to test with hardware as well. 

@adamrankin Could you help with merging this into `master`?

PR #1164, opened 2024-02-27
BUG: Fix handling of timestamp for WinProbe rendered frames

jamesobutler:
This includes a fix for better handling of the timestamp for rendered WinProbe frames because currently the timestamp offset sometimes gets randomly pushed to the future resulting in new frames getting ignored as they are considered "old" resulting in the stream appearing frozen. With these changes I have no longer observed a spurious frame increasing the timestamp so that when the ultrasound sequencer restarts, new frames always come through rather than the occasional longer than expected freeze of the live stream as we wait for new frames to have a timestamp beyond the spurious far one.

cc: @nhjohnston as another person who has used hardware to confirm these changes.


nhjohnston (commented):
Can confirm this fixes the unexpected freezing of the livestream on hardware

jamesobutler (commented):
@Sunderlandkyl or @adamrankin Can you help with the integration of this PR? Thanks!

PR #1173, opened 2024-03-18
BUG: Fix WinProbe live stream not recovering upon depth change

jamesobutler:
Changing imaging depth was triggering start recording which was resetting the frame number and resulting in the live stream not streaming.

In general, changing of the scan depth does not require stopping and starting the recording to execute this parameter change.

cc: @nhjohnston for additional review with hardware


nhjohnston:
I can confirm that this works as expected on hardware 👍

jamesobutler:
@Sunderlandkyl or @adamrankin Can you help with the integration of this PR? Thanks!

PR #1231, opened 2025-06-04
Add support for Clarius Cast v12

auneri:
Addresses connection error caused by a change in CusConnectFn, which now expects imuPort as its second argument.

Note that this change breaks compatibility with <v12. That said, considering that Clarius remotely pushes their updates, I'm not sure if it is necessary to maintain support for older versions.


Sunderlandkyl:
Thanks! No reason to support earlier versions of the Cast device considering we would be forced to update the probe anyway.

PR #1241, opened 2025-07-07
CI: Update ITK version to 5.4.4

jamesobutler:
This updates ITK to use in Plus build CI jobs.

This corresponds to the change in https://github.com/PlusToolkit/PlusBuild/pull/96. @Sunderlandkyl Should these 2 places (Pluslib CI and the PlusBuild) be kept in sync?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         build_type: [Release]
         arch: [x64]
     with:
-      itk-hash: 898def645183e6a2d3293058ade451ec416c4514 # v5.4.2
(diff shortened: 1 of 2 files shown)


Sunderlandkyl:
> Looks like PRs triggered by a ThirdParty that does not have access to https://github.com/PerkLab/PLTools will have the CI jobs fail trying to build Plus as seen in https://github.com/PlusToolkit/PlusLib/actions/runs/16105620098/job/45441082910?pr=1241.

Yeah, in retrospect, that makes sense. Since we build all of the nightly packages including PLTools devices in the nighly task, we can skip the other ones for the ci. I'll update the action to remove PLTools devices.

PR #1243, opened 2025-07-08
ENH: Reworked AtracsysMarkerCreator

atracsys-sbt:
One can easily create an Atracsys geometry file with automatic background subtraction and geometry refinement. It does not require to track an initial guess of the geometry. The geometry layout is consistent and generate simpler geometries:
- The origin is the centroid C of all fiducials
- The x-axis is defined by the largest unique distance from the origin
- The y-axis is in the plane defined by the origin, the x-axis and the furthest fiducial from the x-axis
- The z-axis naturally comes from x and y-axes

This resulted in various changes in the wrapper:
- Added a namespace Atracsys for classes AtracsysTracker (now Tracker), Marker and Fiducial
- Added a namespace Geometry for IniFile class and related free functions
- Added GetFiducialsInFrame function, which allows to retrieve 3D fiducials in the scene without tracking a specific geometry

### DIFF
--- a/src/PlusDataCollection/Atracsys/AtracsysMarkerCreator.cxx
+++ b/src/PlusDataCollection/Atracsys/AtracsysMarkerCreator.cxx
@@ -9,9 +9,12 @@
 #include <iostream>
 #include <vector>
 
+//----------------------------------------------------------------------------
+using namespace Atracsys;
+
 // global handle to Atracsys API wrapper
-AtracsysTracker Tracker;
-AtracsysTracker::DEVICE_TYPE DeviceType = AtracsysTracker::UNKNOWN_DEVICE;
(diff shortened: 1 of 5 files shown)


Sunderlandkyl:
Thanks @atracsys-sbt it all looks good! Will we need to update the SDK to compile the device?

atracsys-sbt:
You don't need to update the SDK, because I have not made much changes to the actual wrapper. That said, I would strongly recommend that you use our latest SDK for your CI so that users would get the latest runtime libraries in PlusToolkit. What versions of the SDK (fusiontrack and sprytrack) do you have ?

Sunderlandkyl:
Looks like fusionTrack is 4.8.1 and spryTrack is 4.8.0.

PR #1281, opened 2026-07-06
Qualify unqualified cout/endl in vtkPlusPolydataForce

hjmjohnson:
`vtkPlusPolydataForce.cxx` used bare `cout`/`endl`, which no longer resolve under current libc++ (`use of undeclared identifier 'cout'`). Add a `using` directive so it compiles. Independent of ITK/VTK version.

<details>
<summary>Details</summary>

Surfaced while building PlusLib against ITK 6 with a recent clang/libc++, but the cause is standard-library strictness, not ITK. One of a set of independent ITK-6 compatibility PRs.
</details>


hjmjohnson:
@lassoan Compiler warnings resolved.

PR #1292, opened 2026-08-26
Fix two Linux build failures: shared StealthLink library and unused vtkImageViewer2 include

### LINKED ISSUE #957: PlusLib build issues with StealthLink
Hi,

I am having issues using PlusBuild. I am using the following cmake settings:
```bash
cmake ../PlusBuild \
  -DSTEALTHLINK_ROOT_DIR=/StealthPyCLinkLibs \
  -DPLUS_USE_STEALTHLINK=ON \
  -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0" \
  -DBUILD_TESTING=OFF \
  -DCMAKE_BUILD_TYPE=Debug
```

I get these errors:
```bash
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusDataCollection.so: undefined reference to `MNavStealthLink::StealthServer::StealthServer(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusDataCollection.so: undefined reference to `MNavStealthLink::Exam::getExamData(MNavStealthLink::StealthServer&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, MNavStealthLink::Error&)'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusDataCollection.so: undefined reference to `boost::system::system_category()'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusDataCollection.so: undefined reference to `MNavStealthLink::StealthServer::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, MNavStealthLink::DataItem&, MNavStealthLink::DateTime&, MNavStealthLink::Error&)'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusDataCollection.so: undefined reference to `boost::system::generic_category()'
collect2: error: ld returned 1 exit status
make[2]: *** [src/PlusDataCollection/CMakeFiles/PlusVersion.dir/build.make:291: /home/mkomaiha/opt/PlusBuild-bin/bin/PlusVersion] Error 1
make[1]: *** [CMakeFiles/Makefile2:1677: src/PlusDataCollection/CMakeFiles/PlusVersion.dir/all] Error 2
```

I think it has to do with "-D_GLIBCXX_USE_CXX11_ABI=0" not getting passed to the PlusLib cmake file. However, when I manually add it to the CXX flags in PlusLib-bin and try to make again I get different issues:

```bash
/bin/ld: CMakeFiles/EditSequenceFile.dir/Tools/EditSequenceFile.cxx.o: in function `MixTrackedFrameLists(vtkIGSIOTrackedFrameList*, std::vector<std::string, std::allocator<std::string> >)':
/home/mkomaiha/opt/PlusBuild-bin/PlusLib/src/PlusCommon/Tools/EditSequenceFile.cxx:147: undefined reference to `igsioTrackedFrame::GetCustomFields()'
...
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusCommon.so: undefined reference to `vtksys::SystemTools::GetCurrentDateTime(char const*)'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusCommon.so: undefined reference to `vtksys::SystemTools::FileIsFullPath(std::string const&)'
/bin/ld: /home/mkomaiha/opt/PlusBuild-bin/bin/libvtkPlusCommon.so: undefined reference to `vtksys::SystemTools::GetCurrentWorkingDirectory(bool)'
collect2: error: ld returned 1 exit status
make[2]: *** [src/PlusCommon/CMakeFiles/EditSequenceFile.dir/build.make:283: /home/mkomaiha/opt/PlusBuild-bin/bin/EditSequenceFile] Error 1
make[1]: *** [CMakeFiles/Makefile2:715: src/PlusCommon/CMakeFiles/EditSequenceFile.dir/all] Error 2
```

Any help would be much appreciated!

mkomaiha:
Two independent build fixes found while building PlusLib with `PLUS_USE_STEALTHLINK=ON` on Linux. They are in separate commits and touch different parts of `src/PlusDataCollection/CMakeLists.txt`, so either can be cherry-picked on its own.

---

## 1. Link the shared StealthLink library on Linux — `BUG:` commit

Addresses part of #957.

On Linux `PlusDataCollection` linked `libStealthLink.a`. A static archive records no dependencies of its own, so the boost symbols the SDK references are never resolved and linking PlusLib fails:

```
undefined reference to `boost::system::system_category()'
undefined reference to `boost::system::generic_category()'
```

The SDK ships its own boost 1.49 beside the library, but nothing tells the linker to pull it in. `libStealthLink.so` records them as `DT_NEEDED`:

```
$ objdump -p libStealthLink.so | grep NEEDED | grep boost
  NEEDED  libboost_signals-mt.so.1.49.0
  NEEDED  libboost_system-mt.so.1.49.0
  NEEDED  libboost_thread-mt.so.1.49.0
```

so linking the shared library resolves them without enumerating the bundled boost in CMake and hardcoding versions that would go stale with the next SDK drop.

The shared library is also added to `External_Libraries_Install`, matching what the `WIN32` branch already did. Both platforms now install the same file, so that call is hoisted out of the conditional and only the genuinely platform-specific part — which library to link, plus the Windows-only debug variant — stays inside it.

**This does not close #957 by itself.** The remaining errors in that issue are ABI mismatches (`std::__cxx11::basic_string` vs `std::string`) caused by `-D_GLIBCXX_USE_CXX11_ABI=0` not reaching the external projects. The SDK is built against the pre-GCC5 libstdc++ ABI, and VTK, ITK, IGSIO and OpenIGTLink all have to agree with it. That fix belongs in PlusBuild and is filed separately; this change is independent of it and correct on its own.

## 2. Remove unused vtkImageViewer2 include from ViewSequenceFile — `COMP:` commit

Unrelated to StealthLink; it just happens to block the same build.

`Tools/ViewSequenceFile.cxx` includes `vtkImageViewer2.h` but never uses `vtkImageViewer2`. The include is the only mention of the name in the file, and always has been — no commit in the file's history has ever referenced the type. The tool builds its display from `vtkImageActor`, `vtkRenderer` and `vtkRenderWindow` instead.

The header belongs to VTK's `InteractionImage` module, which `ViewSequenceFile` does not link, so the include only resolves if that module's include directory happens to arrive through some other VTK target. When it doesn't, the build fails:

```
ViewSequenceFile.cxx:17:10: fatal error: vtkImageViewer2.h: No such file or directory
```

This shows up with VTK 9.1.0 used from a build tree, where include directories come only from the modules actually linked.

Deleting the include is the fix. Linking `InteractionImage` would also make it compile, but the binary references no symbol from that module — `--as-needed` drops it right back out, and the executable is byte-for-byte identical either way.

For contrast, the 14 files that genuinely use this class (all under `Testing/`) include it, declare a `vtkImageViewer2*`, and construct one — and `Testing/CMakeLists.txt` links `InteractionImage` for exactly those targets. `ViewSequenceFile` has only the include.

---

## Testing

Ubuntu 21.04, GCC 10, VTK 9.1.0 and ITK 5.4.4 from a PlusBuild superbuild, StealthLink SDK 2.4, `PLUSBUILD_BUILD_PlusLib_TOOLS=ON`.

Each failure reproduces without its commit, and each commit was verified on its own. With both applied the build completes and the device is available at runtime:

```
$ objdump -p libvtkPlusDataCollection.so | grep NEEDED | grep -i stealth
  NEEDED   libStealthLink.so.2.4.20150414

$ PlusVersion | grep -i stealth
  - StealthLinkTracker (ver: Plus-2.9.0)
```

(The StealthLink half additionally requires the PlusBuild ABI fix mentioned above to reach a successful link.)


adamrankin:
Do we know someone with a StealthLink and Linux to test the fix?

